PR #2, opened 2015-09-29
Add support of official Rapsberry Pi touchscreen (kernel 4.1.8)

ThomArmax:
I have added the support of the official rapsberry pi touchscreen by changing `PREFERRED_VERSION_linux-raspberrypi` to `4.1.%` and adding the associated kernel recipe.

I'm not shure it is the best way to do it, I'm quite new in Yocto and BSP dev.

### DIFF
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -17,12 +17,7 @@ XSERVER = " \
     xf86-video-fbdev \
     "
 
-# Really supported starting from linux-raspberrypi 3.18.y only
-KERNEL_DEVICETREE ?= " \
-    bcm2708-rpi-b.dtb \
-    bcm2708-rpi-b-plus.dtb \
-    bcm2709-rpi-2-b.dtb \
-    \
+KERNEL_DEVICETREE_3.18.y = " \
     ds1307-rtc-overlay.dtb \
     hifiberry-amp-overlay.dtb \
     hifiberry-dac-overlay.dtb \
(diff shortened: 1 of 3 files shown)


agherzan:
As well, if I don't ask too much, could you please read and follow these (http://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines) guidelines for commit log?

ThomArmax:
Hi @agherzan ,
So what's the best way to do it (I'm new in open source contribution, git, and yocto/kernel dev)? Do I have to create a new pull request with clean commits (by a reset of my master) ?
Do you think a solution for 
> This won't work on other intermediate versions and implementing a way to handle it thick defeats the purpose a little bit. So I'm wondering if using only the default rpi kernel version would be the right way to do in order to avoid complexity.

could be to remove the kernel 3.18 and only keep 4.1 version ? This should remove all the complexity.

Thanks by advance

agherzan:
Hi @ThomArmax . I put together a cleanup patch series. Could you please test the screen on them? The branch is ag_k4 on github here.

ThomArmax:
Hi @agherzan , ok thanks. I will test that this evening

ThomArmax:
The screen is recognised and seems to works, but the display is reversed.
I think it's beacause of the missing `rpi-ft5406-overlay.dtb` in overlays

agherzan:
I added in the branch a commit for all the overlays. Could you please
retest now? Make sure you actually activate the overlay in config.txt

--
Andrei Gherzan

ThomArmax:
I still have the same issue, even by adding `dtoverlay=rpi-ft5406` in my `config.txt`.
I've read an article (https://www.raspberrypi.org/forums/viewtopic.php?f=108&t=121263) which says : 
> The Big Problem with all earliest Display Cases is that the Cable-Outputs are on the wrong side, because all manufactures had build their Cases with the 'wrong' Rotation.
Last Week the Rotation changed in Firmware so its now 180° upside down... (i know it was before flipped and now correct, but that doesnt change the Case-Problem)
I think its better to wait a little bit till new Versions of the Display-Cases will arrive

Because I bought the screen and its case in the first days, I have this issue. The kernel rev I used in my meta was 'not fixed' with the screen rotation so it works fine on my branch. Since your kernel rev in your branch is up-to-date I'm faced to this issue, which actually is not.

So I think we can say thats ok on your branch, and on my side, I've to find a solution like https://github.com/remonlam/rpi-touch-display-fix

Thanks.

agherzan:
Is the touchscreen working too? With or without the overlayer?

ThomArmax:
I've not tried yet, a lot of work those days. I'll check that as soon as possible.

agherzan:
Thank you Thomas.

ThomArmax:
Hi @agherzan ,
I'm glad to tell you the touchscreen is working. Test conditions : 
* Image : `qt5-image-demo` ([meta-embarcados](https://bitbucket.org/embarcados/meta-embarcados/src))
* Branch : `ag_k4`
* config.txt : `dtoverlay=rpi-ft5406`
* Application launched: `/usr/share/qt5everywheredemo-1.0/QtDemo -platform eglfs`

If you don't enable the `rpi-ft5406` overlay, the screen works but **not** the touchscreen

agherzan:
Good news @ThomArmax . All we have to do now is figure out how to manage the different dtbs per kernel version. Cause for example rpi-ft5406 is only available in 4.1.x so adding it unconditionally will break older kernels.

PR #13, opened 2016-04-13
RFC: Add preempt-rt recipe for rpi

giraldeau:
*** RFC - DO NOT MERGE ***

The added recipe uses kernel sources including preempt-rt patches and
raspberrypi changes.

However, the kernel configuration is not set correctly. Neither the rt.cfg nor
the do_configure_prepend() method worked. I'm clueless about the proper way to
modify the .config file.

Also, maybe can we merge the dts patch for pitft22 to avoid the duplicated patch?

Signed-off-by: Francis Giraldeau <francis.giraldeau@gmail.com>


giraldeau:
Will open an issue instead.

PR #25, opened 2016-07-29
Patch to work around kernel config not working.

### LINKED ISSUE #14: Extends kernel configuration
The bitbake build process configures the kernel using .cfg files. However, the meta-raspberrypi linux-rpi.inc creates a config file and calls make oldconfig. Is there a reason why the configuration is done this way? What would be the prefered way to extends the kernel configuration with additional parameters?

Thanks!

izak:
This simply applies [this patch](https://lists.yoctoproject.org/pipermail/yocto/2015-August/026101.html) and fixes #14 .

I need this support to enable a touch screen module that's patched in for the Raspberry Pi.

### DIFF
--- a/recipes-kernel/linux/linux-raspberrypi.inc
+++ b/recipes-kernel/linux/linux-raspberrypi.inc
@@ -6,18 +6,15 @@ SECTION = "kernel"
 LICENSE = "GPLv2"
 LIC_FILES_CHKSUM = "file://COPYING;md5=d7810fab7487fb0aad327b76f1be7cd7"
 
-SRC_URI += " \
-        file://defconfig \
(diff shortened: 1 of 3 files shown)


betafive:
This needs a clearly commit message to explain what is being changed. It also needs updating to follow the addition of raspberrypi3-64 support.

agherzan:
@izak I know this is an old contribution but because we haven't used github before, we didn't actually work on this. Would you mind updating the PR and make sure you take a look on http://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines ?

agherzan:
I will close this PR for now.

PR #28, opened 2016-09-21
Fix issue tracker link in README

ali1234:


README
```diff
@@ -5,7 +5,7 @@ Git repository web frontend:
 Mailing list (yocto mailing list):
     yocto@yoctoproject.org
 Issues management (redmine):
```
agherzan: Replace redmine references too. There is one in the content too.

PR #33, opened 2016-10-28
disable failing append to bundle_initramfs task

lewicki-pk:
Hi Andrei,
I'm not totally sure what this append is suppose to do, but I'm sure that "mkknlimg" tool is not in the ${STAGING_BINDIR_NATIVE} (I also think that it's not in the tools anymore).

I have built an image that is using initramfs and it is working with those lines removed.
Without removing them it fails on this step.

### DIFF
--- a/recipes-kernel/linux/linux-raspberrypi.inc
+++ b/recipes-kernel/linux/linux-raspberrypi.inc
@@ -54,16 +54,3 @@ do_deploy_append() {
     fi
     echo "${CMDLINE}${PITFT_PARAMS}" > ${DEPLOYDIR}/bcm2835-bootfiles/cmdline.txt
 }
-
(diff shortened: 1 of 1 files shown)


lewicki-pk:
Ok. I have finally took a deeper look at this issue and it turns out that only  directory is wrong. I'm going to fix this and apply patch on the yocto list- closing this one..

agherzan:
Hello @lewicki-pk . I'm not sure I understand . What directory are you talking about?

lewicki-pk:
Previously I thought that the whole "append" for this task should be removed but now I checked and mkknlimg is in different place. I created a new pull request #34 .
Sorry for confusion.

agherzan:
Uh Right. It makes sense now @lewicki-pk . Thanks for this.

PR #34, opened 2016-10-31
linux-raspberrypi: Fix path for mkknlimg in bundle_initramfs task

lewicki-pk:
Task do_bundle_initramfs() was failing because of "mkknlimg: not found"
The path to "mkknlimg" is now updated.


agherzan:
Actually Your first PR was the right one. Checking the code base I saw that mkknlimg is not needed anymore: see commit fe92cc7e1428ba7f982c371101150851c621532a .

lewicki-pk:
I see. I'll be happy either way if I can use official repo instead of forks :-)

agherzan:
Sounds good.

PR #38, opened 2016-11-24
Do not rely on DISTRO_TYPE to enable/disable kernel debug

fmeerkoetter:
DISTRO_TYPE doesn't seem to be set anywhere in OE/poky. This causes
CMDLINE_DEBUG to default to "debug" which is very noisy. This is
especially true when building a systemd based system, as systemd
also looks at the kernel commandline to set the debug level.
Such a system is nearly unuseable from the serial console due
to the amount of data logged by systemd.


betafive:
@agherzan This is merged, we can close it here.

agherzan:
Correct. Closing.

PR #39, opened 2016-11-26
Split '-' as '.'

Gerharddc:
Treats a '-' character in a kernel version as a '.' when splitting as to support something like 4.9-rc6.


betafive:
@agherzan This is now out-of-date and isn't needed for morty or master. We should close this pull request.

agherzan:
We removed this routine from morty and master indeed.

PR #54, opened 2017-04-21
modify to use xf86-input-libinput

### LINKED ISSUE #45: mouse does not worked at core-image-sato in morty
Hello.

I tried to use core-image-sato with vc4graphics.
Sato is booted however mouse pointer does not moved in spite of mouse is moved.

I added the following line to my local.conf. 
It seems to work fine.

```txt
IMAGE_INSTALL_append = " xf86-input-evdev"
```

I think xf86-input-evdev should be included IMAGE_INSTALL or dependencies to core-image-sato or else...

Would you let me know your opinion.
Thanks.

mickey-happygolucky:
To fix #45 cherry-pick the 52bec45 and adds a bbeppend xserver-xorg which includes xf86-input-libinput in XSERVER_RRECOMMENDS.


betafive:
I'd recommend requesting a backport of the relevant oe-core commit (2d005faf) to morty instead of just fixing it for meta-raspberrypi. Then we should look at backporting the meta-raspberrypi commit 52bec45b. 

@agherzan I think we should close this, I wouldn't like to merge to morty as-is.

agherzan:
For this specific case @betafive , if fixed here or in morty it would not have any impact. I mean they would not conflict. And it's just a simple XSERVER_RRECOMMEND . I generally agree with you and I think that this is in it's core a non BSP issue.

agherzan:
test this please

betafive:
Ok cool, that makes sense. It'd be ideal to fix this in oe-core so it applies to other hardware but let's just get it fixed here for now.

agherzan:
@mickey-happygolucky Paul has a very good point here. Would you want to push the fix in poky too? We will fix it anyway here so you don't have to wait for it.

agherzan:
I checked the morty branch in poky - this fix is not there. Anyway - I'll fix it here for now. Merging.

PR #59, opened 2017-04-28
Make CMDLINE overridable from outside

LocutusOfBorg:
currently there is no way to override the default CMDLINE, lets set if only if not already set from outside

### DIFF
--- a/recipes-kernel/linux/linux-raspberrypi.inc
+++ b/recipes-kernel/linux/linux-raspberrypi.inc
@@ -21,7 +21,7 @@ KERNEL_DEFCONFIG_raspberrypi2 ?= "bcm2709_defconfig"
 KERNEL_DEFCONFIG_raspberrypi3-64 ?= "bcmrpi3_defconfig"
 
 # CMDLINE for raspberrypi
-CMDLINE = "dwc_otg.lpm_enable=0 console=serial0,115200 root=/dev/mmcblk0p2 rootfstype=ext4 rootwait"
+CMDLINE ?= "dwc_otg.lpm_enable=0 console=serial0,115200 root=/dev/mmcblk0p2 rootfstype=ext4 rootwait"
 
 # Add the kernel debugger over console kernel command line option if enabled
 CMDLINE_append = ' ${@base_conditional("ENABLE_KGDB", "1", "kgdboc=serial0,115200", "", d)}'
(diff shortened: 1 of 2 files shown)


LocutusOfBorg:
Can you please also fix for morty and other branches? :)

mirzak:
One could actually override with with an .bbappend file, but I support this change an was actually thinking about sending a patch as well. 

One problem with this, patches should go to the mailing list. See README file in meta-raspberrypi

LocutusOfBorg:
I already have a bbappend, but the override must have ":=" (I think) and this seems somewhat strange

LocutusOfBorg:
thanks I think I forwarded it to the ML!

betafive:
@agherzan Just spotted this on the mailing list, I think we can close this pull request.

recipes-kernel/linux/linux-raspberrypi.inc
```diff
@@ -21,7 +21,7 @@ KERNEL_DEFCONFIG_raspberrypi2 ?= "bcm2709_defconfig"
 KERNEL_DEFCONFIG_raspberrypi3-64 ?= "bcmrpi3_defconfig"
```
agherzan: Can you please modify the commit log in accordance to http://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines ? Looks good to me other than that.

LocutusOfBorg: I think it is good now :)

agherzan:
With one exception : the subject line. It needs to be in the following format : "component : Short description". 

Thanks once again for your contribution.

LocutusOfBorg:
I did correct that subject line and added a new commit on raspberrypi0.conf :)

agherzan:
Hi @LocutusOfBorg I'm really sorry for buzzing you again. But I think once you get it, everything will be just fine.

The first commit makes a change in the linux-raspberrypi component. The commit subject therefore should have a format like: "linux-raspberrypi: Make CMDLINE overridable from outside".

Similar for the second commit where the component is a machine configuration: "raspberrypi0.conf: make SERIAL_CONSOLE overridable from outside"

Take a look [here](http://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines) for additional information.

agherzan:
ok to test

LocutusOfBorg:
thanks! :) I guess I fixed it correctly

agherzan:
@LocutusOfBorg looks good now. Thanks for this.

LocutusOfBorg:
thanks for teaching me how to fix the commit messages :)

agherzan:
With pleasure. Merging this.

PR #84, opened 2017-06-13
bluez5: install patches for all rpi models

michal-mazurek:
This is required to make rpi0-wireless bluetooth work.

### DIFF
--- a/recipes-connectivity/bluez5/bluez5_%.bbappend
+++ b/recipes-connectivity/bluez5/bluez5_%.bbappend
@@ -1,6 +1,6 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 
-SRC_URI_append_raspberrypi3 = " \
+SRC_URI_append = " \
     file://BCM43430A1.hcd \
     file://0001-bcm43xx-Add-bcm43xx-3wire-variant.patch \
     file://0002-bcm43xx-The-UART-speed-must-be-reset-after-the-firmw.patch \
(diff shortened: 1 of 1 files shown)


agherzan:
I think this PR raises a problem - RaspberryPi Zero WiFi doesn't include all the bits needed for bluetooth. We can:
1. Add the patches and firmware for any rpi (which will mean rpi2 will include stuff it doesn't need but I see it as a tiny side effect)
2. Use a new SOC_FAMILY `rpiwifi` for example in witch rpi3 and rpi0wifi will be included and these bluez5 modifications will be done based on it.

@betafive @kraj What do you guys think?

betafive:
If this is contained to "bluez5_%.bbappend" then there should be no impact on rpi devices which don't install bluez5 and a minor waste on those that do install bluez5 but don't need that firmware. So if it's contained to that file then I'd just apply the changes for all rpi devices.

agherzan:
It is indeed contained right now in bluez5 package. That's why I consider it as a good compromise.

agherzan:
@michal-mazurek could you push a change to this which uses `_rpi` so that we apply these change to all the rpi machines?

recipes-connectivity/bluez5/bluez5_%.bbappend
```diff
@@ -1,6 +1,6 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 
-SRC_URI_append_raspberrypi3 = " \
+SRC_URI_append = " \
```
kraj: if they should apply to all rpi SOCs then use `_rpi` override please. We can not add patches unconditionally since this will mess up other BSPs when doing multi machine builds.

agherzan:
@kraj :) see the comments above.

PR #88, opened 2017-06-27
raspberrypi3-64.conf: Don't use raspberrypi as MACHINEOVERRIDES

michal-mazurek:
The current setup broke the build for rpi3-64 when we wanted to port
some changes from rpi3 to rpi0.

### DIFF
--- a/conf/machine/raspberrypi3-64.conf
+++ b/conf/machine/raspberrypi3-64.conf
@@ -2,7 +2,7 @@
 #@NAME: RaspberryPi 3 Development Board
 #@DESCRIPTION: Machine configuration for the RaspberryPi 3 in 64 bits mode
 
-MACHINEOVERRIDES = "raspberrypi3:raspberrypi:${MACHINE}"
+MACHINEOVERRIDES = "raspberrypi3:${MACHINE}"
 
 MACHINE_EXTRA_RRECOMMENDS += "linux-firmware-bcm43430"
 


agherzan:
@michal-mazurek could you add to this PR the change for bluez too?

agherzan:
Actually that was not merged. Wouldn't that help you too?

agherzan:
Friendly ping towards @michal-mazurek .

michal-mazurek:
I'll get back to this PR soon.

agherzan:
@michal-mazurek any update on this?

michal-mazurek:
Did you mean this commit? https://github.com/resin-os/resin-raspberrypi/commit/91e567ee290501311897bce92f6f627c2335c480 -- do you think it is still necessary after the change from this commit is applied?

agherzan:
I'm a little confused @michal-mazurek . That is in resinOS. I was talking about this commit in the BSP.

michal-mazurek:
You asked me to "add to this PR the change for bluez". I was asking which "change for bluez" specifically you meant. Is it the one I linked to?

conf/machine/raspberrypi3-64.conf
```diff
@@ -2,7 +2,7 @@
 #@NAME: RaspberryPi 3 Development Board
 #@DESCRIPTION: Machine configuration for the RaspberryPi 3 in 64 bits mode
 
-MACHINEOVERRIDES = "raspberrypi3:raspberrypi:${MACHINE}"
+MACHINEOVERRIDES = "raspberrypi3:${MACHINE}"
```
kraj: It needs some more explanation. raspberrypi is a SOC family override and it natually would be used on all machines based on rpi. if its conflicting then we need to find a portable way to apply fixes for both pi0 and pi3-64

agherzan: The soc family is actually `rpi` and not `raspberrypi`.

kraj: I realized that, but then raspberrypi is confusing, it is mean for rpi1 ?

agherzan: Yup. That's the confusing part. But calling it raspberrypi1 might be pretty strange too.

kraj: It wont be since it falls in with the naming conventions that rpi boards have used but then they did not do it for the first board where 1 is sort of implicit. 

in anycase, I am fine with removing this rpi1 override from rpi3_64 machine overrides. Although it needs to go along with building and booting a rpi3_64 image

agherzan: Sure indeed. Anyway this was only for morty. We already have this change in master (and pyro afaik).

kraj: OK I did not realize it was a cherry-pick, it sounded like a complete new patch

PR #96, opened 2017-07-14
userland: Add missing EGL_CAST defines

gizero:
Recent updates of libepoxy in oe-core broke the build for raspberrypi of dependent packages like gtk+3, due to a missing definition in an glplatform.h header provided by userland.
This patch backports the missing define from the updated version of the header available in mesa, hence restoring successful builds.
The patch has been submitted upstream here: https://github.com/raspberrypi/userland/pull/407

See http://lists.openembedded.org/pipermail/openembedded-core/2017-July/139311.html for relevant discussions.


agherzan:
ok to test

Issue #105, opened 2017-08-03
wiringpi library is not linked correctly


aurelian17:
**Description**

Incorrect order of compiler flags causing the libwiringpi.so library to not have the required linking dependencies.

**Steps to reproduce the issue:**
Build an executable that links to libwiringpi.so. The compilation of this executable will break with the following error:
recipe-sysroot/usr/lib/libwiringPi.so: undefined reference to `crypt'
| collect2: error: ld returned 1 exit status

**Describe the results you received:**
readelf -d libwiringPi.so

Dynamic section at offset 0xbf18 contains 25 entries:
  Tag        Type                         Name/Value
 0x00000001 (NEEDED)                     Shared library: [libc.so.6]
 0x0000000e (SONAME)                     Library soname: [libwiringPi.so.]

**Describe the results you expected:**
readelf -d libwiringPi.so

Dynamic section at offset 0xbef8 contains 29 entries:
  Tag        Type                         Name/Value
 0x00000001 (NEEDED)                     Shared library: [libm.so.6]
 0x00000001 (NEEDED)                     Shared library: [libpthread.so.0]
 0x00000001 (NEEDED)                     Shared library: [librt.so.1]
 0x00000001 (NEEDED)                     Shared library: [libcrypt.so.1]
 0x00000001 (NEEDED)                     Shared library: [libc.so.6]
 0x0000000e (SONAME)                     Library soname: [libwiringPi.so.]

**Additional information:**
I reproduced the problem on pyro branch, but I'm quite sure that can be reproduced on master too. It's 100% reproducible. I already have the fix and I'll submit it as soon as possible.

PR #109, opened 2017-08-20
Fix for wiringpi library that is not linked correctly

aurelian17:
Incorrect order of compiler flags causing the libwiringpi.so library
to not have the required linking dependencies. The cross-compilation
patch needed to be reworked in order to fix issue #105.

Signed-off-by: Aurelian Zanoschi <aurelian17@gmail.com>

**- How I did it:**
The major problems were in the Makefile from wiringPi source tree that is currently patched by 0001-Add-initial-cross-compile-support.patch. I had to fix the problems and rework the patch.


agherzan:
add to whitelist

agherzan:
test this please

agherzan:
Thanks @aurelian17 for the fix. Could you please tweak the commit log so that it would keep the format used in this repos (http://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines)?

agherzan:
There is though one more change needed. Builds got broken due to a bbappend getting invalid. This was fixed in master. Could you please rebase this PR for the build server to be able to pass the issue currently fixed in master?

aurelian17:
Andrei, can you please push this commit to the stable pyro branch too? Unfortunately that stable branch is affected by this problem too.

Thanks in advance.

gizero:
Was there any attempt to talk to the upstream to get the content of the original/updated patch integrated? It looks like it deals with very generic cross-compile related issues. I had to dig into it months ago while upgrading wiringPi to 2.44 and I forgot what a mess it was. Now, trying to review what was changed by this patch brought back that nasty feeling...

agherzan:
I was surprised to be honest that it didn't happen. @gizero and/or @aurelian17 can any of you work with upstream to fix this?

@aurelian17 you will need to push a PR towards pyro too in order to have it there.

gizero:
At first sight the upstream project's website doesn't even seem to have a contribution guide. Googling for "site:wiringpi.com patch bug contribute" gives no match! :-() Official sources are on a self hosted git repo with no issue tracker I could find. It might be an harder than expected task :-(. Let's see what's @aurelian17's opinion on that..

aurelian17:
Hi, sorry for the delay, just returned from vacation. From my point of view is quite difficult to push this kind of patches to Gordon's repository.

Thanks, @agherzan for accepting this change, I will push it to the pyro branch too.

PR #110, opened 2017-08-20
Fix issue 14 - extends kernel configuration

andig75:
This patch fixes issues #14 

I have updated the patch available in the mailing list and with this patch is possible to extend the kernel configuration via bbappend and cfg fragments.


agherzan:
It looks good. Could you please fix the commit log accordingly to http://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines . Something like "linux-raspberrypi: Fix kernel configuration using cfg files" and use a recommended line size of 78 chars.

andig75:
@agherzan I have changed the comment
Please let me know if it is ok

agherzan:
It is just fine. Thanks for this. There is though one more change needed. Builds got broken due to a bbappend getting invalid. This was fixed in master. Could you please rebase this PR for the build server to be able to pass the issue currently fixed in master?

andig75:
@agherzan , please let me know if now it is ok, or if something is still missing

agherzan:
Could you please rebase your branch on top of master and remove the merge branch you curently have? Everything else looks good now.

andig75:
I have just created a new PR based on latest master. 
I close this one. The diff is the same

PR #114, opened 2017-08-25
Update kernel & firmware

betafive:
Both the Linux kernel and the rpi firmware are updated to the latest tags dated 2017-08-11.


betafive:
Boot tested on an original rpi and an rpi3 (32-bit).

PR #115, opened 2017-08-29
Fix issue 14 - extend kernel configuration

andig75:
This patch fixes issues #14 and replace previous PR

I have updated the patch available in the mailing list and with this patch is possible to extend the kernel configuration via bbappend and cfg fragments.


agherzan:
test this please

agherzan:
Merged. Thanks @andig75

PR #118, opened 2017-09-07
Various package updates

agherzan:
**- What I did**
Updated varios packages to latest versions.
Integrate with pypi bbclass when possible.

**- How I did it**
Build tested on RaspberryPi 3


betafive:
I ran into two issues when testing these patches:

* Recipes now don't parse without the meta-python layer present. Previously I could "bitbake rpi-hwup-image" with just poky + meta-raspberrypi. This is acceptable though as we do note the dependency in the README file.

* omxplayer tries to talk to a web service hosted on heroku during the build to generate the man page. See https://github.com/popcornmix/omxplayer/blob/master/Makefile#L69. This makes me cry and breaks offline builds, reproducibility, etc. I've got a patch to avoid this by not building the man page which I'll send as a follow up. I'll also raise this upstream as it's absurd.

agherzan:
@betafive there is actually already a mention on meta-python as being a dependency of this layer : https://github.com/agherzan/meta-raspberrypi/blob/master/README.md#dependencies

I will submit a PR with that commit too. Thanks.

agherzan:
Actually, can you please rebase it and push a PR? That would help.

betafive:
Will do. I've got a whole bunch of kernel/u-boot changes I'm currently testing so I'll include it with those.

PR #120, opened 2017-09-08
Kernel, u-boot and omxplayer improvements

betafive:
* Fix omxplayer build so that we don't need the internet during do_compile.

* Refactor the kernel recipes, merging the two inc files together in the process. Drop duplicate, outdated and incorrect code from the recipes.

* Update linux-raspberrypi-dev to the 4.13.y series.

* Support u-boot on raspberrypi3-64. This requires a change to the method of enabling u-boot as we can no longer just infer this from the KERNEL_IMAGETYPE value. Therefore a new RPI_USE_U_BOOT variable is used.

* Fix some minor issues in sdcard_image-rpi.bbclass found whilst implementing the above.


betafive:
Boot tested on raspberrypi3 (32-bit) and raspberrypi3-64, with and without u-boot in both cases. Everything looks good. On 32-bit I've even confirmed that we can use zImage with u-boot instead of uImage if we want to.

If this is too big to review in one go let me know and I'll split it up further. It's been a productive Friday :)

betafive:
Updated after testing with wic images. It turns out there was some dependency information missing.

recipes-multimedia/omxplayer/omxplayer/0006-Prevent-ffmpeg-configure-compile-race-condition.patch
```diff
@@ -0,0 +1,39 @@
+From 4a13419e3805b541fc58e57f1f27c4a388609ef5 Mon Sep 17 00:00:00 2001
+From: Paul Barker <pbarker@toganlabs.com>
+Date: Thu, 7 Sep 2017 20:02:15 +0000
+Subject: [PATCH] Prevent ffmpeg configure/compile race condition
+
+Additional dependency information is needed in Makefile.ffmpeg to ensure that
+the configure stage is finished before the compile stage starts.
+
+Signed-off-by: Paul Barker <pbarker@toganlabs.com>
+Upstream-status: Inappropriate
```
agherzan: This doesn't sound like an inappropriate patch. Would it make sense to upstream it?

betafive: I was a bit unsure on what to put as Upstream-status here. We should try to upstream a fix for this and I've opened the bug https://github.com/popcornmix/omxplayer/issues/567 to work towards that. However this patch is based on the previous one and can't be upstreamed exactly as-is.

agherzan: This sounds like Pending, doesn't it?

betafive: Ok cool I'll change it to Pending.

betafive:
I've updated the Upstream-status tag and added a comment to explain the use of ffmpeg in the omxplayer recipe as I realised it wasn't immediately obvious.

agherzan:
@betafive could you please rebase this? Looks good.

betafive:
Rebase done, I'm rebuilding this and plan to re-do a couple of boot tests.

betafive:
Rebase done, all now up to date. I'm still having issues with HMDI output on raspberrypi3-64 but that's not related to these changes.

agherzan:
Thanks @betafive . Merged to master.

PR #125, opened 2017-09-14
u-boot: drop now upstreamed patches

gizero:
oe-core now provides v2017.09 of u-boot which already merged both patches
introduced by commit 94e2929f746f7e49a7870f7ea889dcbed05296c7 so we can
drop them from meta-raspberrypi.

PR #126, opened 2017-09-15
u-boot: drop upstreamed patches included in v2017.09

mattsm:
These are now in upstream, so they are no longer required here.

Signed-off-by: Matthew McClintock <msm-oss@mcclintock.net>

**- What I did**

**- How I did it**


gizero:
@mattsm Looks the same patch that I sent in #125

mattsm:
yes, dropping this...

PR #134, opened 2017-10-02
linux-firmware: Update brcmfmac43430 to 7.45.41.46

agherzan:
The version of this firmware provided by linux-firmware is 7.45.41.26. This
version seems to have connectivity issues when bonding interfaces. Raspbian
included an updated version (7.45.41.46) which tested, proved to fix this
problem (see https://github.com/resin-os/meta-resin/issues/812).

The version Raspbian is providing is not yet in linux-firmware as they have it
directly from Cypress (see https://github.com/RPi-Distro/repo/issues/88).
Besides the issue described above, it fixes various other issues including a
broadpwn fix.

Hopefully this version, or a newer one, will end up in upstream soon. In order
to catch when that happens and stop overwriting this firmware, a check is added
in the racipe based on the md5sum of the current shipped firmware.

In terms of ditribution license, we got an informal one in the github thread
mentioned above.

Signed-off-by: Andrei Gherzan <andrei@gherzan.com>

**- What I did**

Backport from master

**- How I did it**


betafive:
Looks good.

I'll pull this into pyro now. Still planning to review recent commits and see what else should be backported, I've been delayed due to lack of internet for the office though. Hopefully that should finally be solved tomorrow though and I can get back to some approximation of normality.

PR #136, opened 2017-10-03
weston: remove rpi-backend configuration switch

fgiancane8:
rpi-backend.so was an attempt to create a specialized
weston backend to be used on raspberry pi like platforms.

At the moment, this backend's support has been dropped in favor
of using the standard drm backend, as the vc4 driver is now mainline
in mesa and in kernel (for kernel support).

As a result, weston on raspberry pi does not require the rpi-backend.so
anymore, nor it is buildable (weston's configure complains about
unrecognized configuration switch).

This patch enables weston to natively run on the DRM backend.

Signed-off-by: Francesco Giancane <francescogiancane8@gmail.com>

**- What I did**

**- How I did it**


fgiancane8:
https://www.phoronix.com/scan.php?page=news_item&px=Weston-Nukes-RPi-Code for the reference informations

agherzan:
ok to test

agherzan:
Thanks @kraj for the second look.

PR #137, opened 2017-10-09
Fix HDMI output issues on raspberrypi3-64

betafive:
These changes fix the HDMI output issues I've been facing with raspberrypi3-64 and a DVI monitor (connected via a HDMI to DVI cable).


betafive:
@kraj, I'd like your input on this as you made some of the original raspberrypi3-64 changes I'm modifying here. Is this likely to cause issues for people using true HDMI monitors with raspberrypi3-64?

PR #158, opened 2017-11-20
rpi-base.inc: Fix KERNEL_IMAGETYPE assignment

betafive:
We can't just override KERNEL_IMAGETYPE in machine-specific conf files without
breaking the implementation of RPI_USE_U_BOOT. Instead we need to define a new
KERNEL_IMAGETYPE_DIRECT variable which will control the value when u-boot is not
in use. This new variable may then be overridden as needed without breaking our
u-boot support.

### DIFF
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -51,7 +51,9 @@ KERNEL_DEVICETREE ?= " \
 #   within u-boot to load the kernel.
 KERNEL_BOOTCMD ??= "bootm"
 KERNEL_IMAGETYPE_UBOOT ??= "uImage"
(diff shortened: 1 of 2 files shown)


betafive:
@shr-project, could you take a look at this potential fix for the issues we've been seeing. I've boot tested on raspberrypi3 and raspberrypi3-64, with and without u-boot enabled in both cases. Builds & boots correctly for all 4 configurations.

It's a little more complex than I would have liked but I don't think there's a way to make things simpler without breaking the way RPI_USE_U_BOOT works.

shr-project:
@betafive It looks good to me. I've replaced my "fix" with this one to see if it will catch some unexpected issue, but it's definitely better than mine.

agherzan:
This looks fine to me. I agree it's a cleaner approach.

PR #163, opened 2017-11-27
Added sample configuration and a new image

FrankBau:
**- What I did**

I have added sample conf files for images with EGL support but no X11. This makes it more clear (at least to me) how the meta-raspberrypi layer could be used.

I have also added a richer audio+video image rpi-av-image demonstrating the EGL support with gstreamer, OpenGL ES and on-target software development.

First tests with this image are described here: https://github.com/FrankBau/raspi-repo-manifest/wiki/regression-testing

The new image contains smaller .mp4 movies (sintel trailer) in a .mp4 container instead of the often used huge bbb movies in meanwhile outdated .avi format.

I'm using this image an my lab and will continue using it in the foreseeable future.

### DIFF
--- a/conf/bblayers.conf.sample
+++ b/conf/bblayers.conf.sample
@@ -1,0 +1,19 @@
+# POKY_BBLAYERS_CONF_VERSION is increased each time build/conf/bblayers.conf
+# changes incompatibly
+POKY_BBLAYERS_CONF_VERSION = "2"
+
+BBPATH = "${TOPDIR}"
+BBFILES ?= ""
+
+BBLAYERS ?= " \
+    ##COREBASE##/meta \
+    ##COREBASE##/meta-poky \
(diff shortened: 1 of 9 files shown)


betafive:
I don't think most of this belongs in a BSP layer sadly.

The sample conf files would suit a distro layer much better. I'm also weary of carrying them in meta-raspberrypi as we'd have to watch the conf files in openembedded-core for changes and merge them in to stay up-to-date.

The sample movie would probably be better in meta-multimedia.

The rpi-av-image makes distro specific decisions like setting NTP servers. Again, we don't really want to do that in a BSP layer.

The packagegroups don't appear to be Raspberry Pi specific.

I think the best thing to do with metadata like this is to start a new distribution layer, define your own distro configuration and use that for your A/V demo.

agherzan:
I share Paul's view on this. So for now, I will close this PR.

PR #164, opened 2017-11-27
Fix KERNEL vars having no effect on kernel recipes.

kacf:
In 4d3a148b8eaca, a new default to KERNEL_IMAGETYPE_UBOOT was added to
raspberrypi3-64.conf, which means that the generic default had to be
changed from '?=' to '??='. However, this presents a problem when
building the kernel recipes, because the '??=' assignment is no longer
strong enough to win over the '?=' which exists in the upstream
kernel.bbclass file. So instead, turn the assignment back to '?=', and
rely on ordering of include files to get correct assignments for
raspberrypi3-64.

connected to #153.

### DIFF
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -49,9 +49,9 @@ KERNEL_DEVICETREE ?= " \
 #
 # * When u-boot is enabled use the "uImage" format and the "bootm" command
 #   within u-boot to load the kernel.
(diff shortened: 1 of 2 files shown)


betafive:
See https://github.com/agherzan/meta-raspberrypi/pull/158

kacf:
Oh nice, even better than this fix!

agherzan:
@kacf The fix is available now in master.

kacf:
@agherzan: Thanks!

PR #176, opened 2018-01-09
rpi-base.inc: use $KERNEL_PACKAGE_NAME for the kernel package name

rsalveti:
The kernel class is now using $KERNEL_PACKAGE_NAME to set the default
kernel package name in order to allow alternate kernel flavors.

This fixes the following bitbake warning:
Variable key RDEPENDS_${KERNEL_PACKAGE_NAME}-base (${KERNEL_PACKAGE_NAME}-image) replaces original key RDEPENDS_kernel-base ().

Signed-off-by: Ricardo Salveti <ricardo@opensourcefoundries.com>


agherzan:
ok to test

PR #180, opened 2018-01-19
linux-firmware: make rpi fw specific to rpi

martingkelly:
Currently, we are installing the rpi custom txt and fw files whenever
the meta-raspberrypi layer is included in bblayers.conf. This breaks
other machines that need to use the bcm43430 drivers but with alternate
txt or fw files.

Add a MACHINE_FEATURE rpi-bcm43430 and make the linux-firmware.bbappend
conditional upon it. This is not pretty, but it does work and I don't
see a cleaner way to do it.

Signed-off-by: Martin Kelly <mkelly@xevo.com>

### DIFF
--- a/recipes-kernel/linux-firmware/linux-firmware_%.bbappend
+++ b/recipes-kernel/linux-firmware/linux-firmware_%.bbappend
@@ -1,26 +1,26 @@
-FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+FILESEXTRAPATHS_prepend_rpi := "${THISDIR}/files:"
 
-SRC_URI += " \
-	file://brcmfmac43430-sdio.bin \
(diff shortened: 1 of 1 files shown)


betafive:
Could we use the custom txt and fw files for all Raspberry Pi models? The change should be harmless for devices without the bcm43430. That would allow us to use the `_rpi` machine override instead of defining a new machine feature.

agherzan:
I vote for simplifying it for now to include this firmware and nvram configuration on all RPi models.

agherzan:
test this please

agherzan:
@martingkelly Yes. This is indeed a limitation. I'll leave it like this for now for the sake of simplicity. If people will find this use-case we can improve.

martingkelly:
Yep, sounds fine to me.

PR #182, opened 2018-01-22
u-boot: drop upstreamed patches in v2018.01

gizero:
These are now in upstream, so they are no longer required here.

Signed-off-by: Andrea Galbusera <gizero@gmail.com>


agherzan:
ok to test

Issue #181, opened 2018-01-22
config.txt customisation overwrites more variables than it should


hhromic:
**Description**
The recipe https://github.com/agherzan/meta-raspberrypi/blob/master/recipes-bsp/bootfiles/rpi-config_git.bb is overwriting variables that it shouldn't.
For example the customisation of `over_voltage` does not only customise the `over_voltage` variable in the config file but also the `over_voltage_min`, `over_voltage_sdram`, `over_voltage_sdram_c`, `over_voltage_sdram_i` and `over_voltage_sdram_p` variables.
This results in `over_voltage` being duplicated in many places in the config file where the other variables with the same initial names are.
This is the config file being fetched by the recipe and then customised: https://github.com/Evilpaul/RPi-config/blob/master/config.txt

Fortunately, the solution to this is simple and straightforward. The sed script in the recipe should include the `=` sign after the variable name, like this:

    # these lines
    sed -i '/#over_voltage/ c\over_voltage=${OVER_VOLTAGE}' ${DEPLOYDIR}/bcm2835-bootfiles/config.txt

    # should be like these lines (note the = sign after the variable name)
    sed -i '/#over_voltage=/ c\over_voltage=${OVER_VOLTAGE}' ${DEPLOYDIR}/bcm2835-bootfiles/config.txt

The above should be done in every `sed` line that customises varaibles.
It's a small bug, harmless, but is a bug anyway and in future could cause problem if other variable names clashes.

**Steps to reproduce the issue:**
1. Just build an image and check the resulting `config.txt` file in `/boot`. You will notice multiple repetitions of `over_voltage` or `cpu_freq`, because of multiple variables beginning with those names.
2.
3.

**Describe the results you received:**
A config.txt file with duplicate variables.

**Describe the results you expected:**
A config.txt file with only the customised variable activated, and other variables with the same initial name left untouched.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Additional details (revisions used, host distro, etc.):**
This in the `master` branch.

hhromic:
thanks for merging the PR!
I confirm it works fine. I'm closing this issue now.

PR #186, opened 2018-01-29
rpi-config: fix config variables with the same initial name being ove…

hhromic:
…rwritten

**- What I did**
I fixed the bug described in Issue #181.
Summary: variables with the same initial name gets repetitions of the original name. For example, if you customise `arm_freq=1000`, the following lines in the original config file:

    #arm_freq=700
    #arm_freq_min=700

Are both replaced with the following repetition:

    arm_freq=1000
    arm_freq=1000

This is because the sed scripts:

    sed -i '/#arm_freq/ c\arm_freq=${ARM_FREQ}' ${DEPLOYDIR}/bcm2835-bootfiles/config.txt

match lines that start with `#arm_freq` instead of the more precise `#arm_freq=`.

**- How I did it**
I added equal (`=`) signs after the variable name matching in the sed scripts. This makes the sed scripts to precisely match the intended variable and not any other with the same starting name.

### DIFF
--- a/recipes-bsp/bootfiles/rpi-config_git.bb
+++ b/recipes-bsp/bootfiles/rpi-config_git.bb
@@ -31,25 +31,25 @@ do_deploy() {
     cp ${S}/config.txt ${DEPLOYDIR}/bcm2835-bootfiles/
 
     if [ -n "${KEY_DECODE_MPG2}" ]; then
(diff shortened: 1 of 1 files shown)


agherzan:
test this please

agherzan:
This looks hood @hhromic . Thanks.

agherzan:
Fixes #181

PR #188, opened 2018-01-31
recipes: use oe.utils.conditional instead of deprecated base_conditional

shr-project:


kraj:
@agherzan this should be merged, all other PR builds will fail otherwise.

gizero:
I think this patch is good and I agree with @kraj it is high priority to merge it, but I also guess PR builds are failing anyway because of similar issues on recipes from meta-openembedded (i.e. https://yocto-ci.resin.io/job/meta-raspberrypi-generic/428/console). Relevant patches by @shr-project are already out on the ML and need to be merged as well to restore PR builds here.

agherzan:
I'm going to merge this without waiting for builds to succeed. They break because meta-oe didn't merge a similar change.

kraj:
I have tested this patch with all other patch to meta-oe etc. and it worked well.

PR #191, opened 2018-02-01
[kronos]wayland lib linking added for libbrcmEGL to avoid undefined symbol while linking libbrcmEGL.so

Haseenamol:
In wayland enabled build libbrcmEGL lib is using wayland functions, but during the build it is not linking with wayland libraries. Hence if we try to link libbrcmEGL, will get undefined symbols. 

I have created patch against userland libbrcmEGL.so creation part to include wayland libraries also int he linking list


kraj:
This patch should be applied after the userland upgrade patch lands.

agherzan:
@Haseenamol Please use the patch guidelines in the documentation and update your git commit message accordingly (and add upstream status to the patch).

Haseenamol:
Hi Khem and Andrei,
  Thank you for the comments.
Andrei,
I will check guideline and make the required changes
Khem,
So that means I have to hold this until userland upgrade patches come, right ?

agherzan:
@Haseenamol The PR Khem was referring to was merged to master. Please rebase yours.

Haseenamol:
Thank you Andrei for the update. I will try to close this by next week

Haseenamol:
We are not getting egl error after upgrading the userlands. So closing this request.
Thank you

agherzan:
Thanks @Haseenamol

PR #192, opened 2018-02-01
omxplayer: make samba support configureable

shr-project:


agherzan:
Looks good. Thanks @shr-project

PR #193, opened 2018-02-05
[linux-firmare] fix addition of NVRAM file to image (fixes #184)

### LINKED ISSUE #184: Switching from rocko to master on rpi-0w breaks wifi
**Description**

Changing from latest on rocko to latest on master makes it so that the `wlan0` interface doesn't show up anymore.

**Steps to reproduce the issue:**
1. pull latest master
2. build `rpi-hwup-image`
3. run `ip link` or anything that lists the interfaces, only `l0` shows up

**Describe the results you received:**
`wlan0` isn't connected

**Describe the results you expected:**
`wlan0` should be present.

**Additional information you deem important (e.g. issue happens only occasionally):**

Switching back to rocko without changing anything makes it work again.

**Additional details (revisions used, host distro, etc.):**
I also tried updating poky to master and it didn't improve anything.
The reason I'm updating to master is to make use of #183

hhromic:
This PR fixes #184, where the Wifi module is not brought up because of missing firmware file.

**- What I did**
The NVRAM text file for the BCM firmware was not properly added in the built image. This is because the recipe `recipes-kernel/linux-firmware/linux-firmware_%.bbappend` is using `+=` for adding the extra file instead of `_append` in `FILES_${PN}-bcm43430`.
This results in only the NVRAM text file being added to the built image and not the firmware binary file (`.bin`), causing the Wifi module to not work.
I tested the images also enabling the bluetooth module (as discussed in #184) to ensure this doesn't interfere with other connectivity components.
After this fix, both files (`.txt` and `.bin`) of the firmware are now correctly installed in `/lib/firmware`.

**- How I did it**
I simply replaced the `+=` construction with an `_append` construction as seen in the PR.

### DIFF
--- a/recipes-kernel/linux-firmware/linux-firmware_%.bbappend
+++ b/recipes-kernel/linux-firmware/linux-firmware_%.bbappend
@@ -21,6 +21,6 @@ do_install_append_rpi() {
     install -m 0644 $_firmware ${WORKDIR}/brcmfmac43430-sdio.txt ${D}${nonarch_base_libdir}/firmware/brcm
 }
 
-FILES_${PN}-bcm43430_rpi += " \
+FILES_${PN}-bcm43430_append_rpi = " \
     ${nonarch_base_libdir}/firmware/brcm/brcmfmac43430-sdio.txt \
 "


hhromic:
Sure, will send an updated PR.

agherzan:
@hhromic You can push the same branch here. No need to create a new PR. Just force push a new commit to the same branch.

hhromic:
Thanks. I created a new PR because when trying to amend the log commit I mistakenly created a merge commit that looks ugly. For this reason I would prefer this clean PR to be merged instead.
Sorry for the annoyance!

agherzan:
That can also be fixed with: cherry pick and or rebase.

hhromic:
I see.  You wish me to attempt to amend this PR or we can just close it and merge the other clean one?
I already deleted the original branch of this PR (in an attempt to start again). So not sure what to do.

agherzan:
Don't worry. Just hints for future.

hhromic:
Thanks for the advice, for sure there is a lot to learn.
Happy to help developing `meta-raspberrypi` !

agherzan:
You're more than welcome.

PR #201, opened 2018-02-27
Kernel updates and other improvements

betafive:
* Fix do_image_rpi_sdimg/do_image_wic dependencies
* rpio: Use pypi.bbclass
* rpi-*-image: Deprecate old image names
* Update kernels to 4.9.80, 4.14.22, 4.15.y
* Update firmware
* Switch default kernel to 4.14.y series

Tested a mix of different distros, images, configurations (with/without u-boot, 4.9.y/4.14.y kernels, etc) and devices (original rpi, rpi2, rpi3, rpi-64). All boots to a login prompt happily :)

### DIFF
--- a/conf/machine/include/rpi-default-versions.inc
+++ b/conf/machine/include/rpi-default-versions.inc
@@ -1,3 +1,3 @@
 # RaspberryPi BSP default versions
 
-PREFERRED_VERSION_linux-raspberrypi ??= "4.9.%"
+PREFERRED_VERSION_linux-raspberrypi ??= "4.14.%"
--- a/recipes-core/images/rpi-basic-image.bb
+++ b/recipes-core/images/rpi-basic-image.bb
@@ -4,3 +4,7 @@ include rpi-hwup-image.bb
 SPLASH = "psplash-raspberrypi"
 
 IMAGE_FEATURES += "ssh-server-dropbear splash"
+
+do_image_prepend() {
+    bb.warn("The image 'rpi-basic-image' is deprecated, please use 'core-image-base' instead")
+}
--- a/recipes-core/images/rpi-basic-image.bb
+++ b/recipes-core/images/rpi-basic-image.bb
@@ -1,5 +1,10 @@
-# Base this image on rpi-hwup-image
-include rpi-hwup-image.bb
+# Base this image on core-image-minimal
+include recipes-core/images/core-image-minimal.bb
+
+# Include modules in rootfs
+IMAGE_INSTALL += " \
+	kernel-modules \
+	"
 
 SPLASH = "psplash-raspberrypi"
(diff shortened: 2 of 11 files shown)


recipes-core/images/rpi-basic-image.bb
```diff
@@ -1,5 +1,10 @@
-# Base this image on rpi-hwup-image
-include rpi-hwup-image.bb
+# Base this image on core-image-minimal
+include recipes-core/images/core-image-minimal.bb
```
kraj: I think it should use core-image-base, which is more usable then core-image-minimal

betafive: I tried to go with minimal change from the current behaviour and rpi-hwup-image is based on core-image-minimal. They're both deprecated now though so I don't think it's worth changing.

conf/machine/include/rpi-default-versions.inc
```diff
@@ -1,3 +1,3 @@
 # RaspberryPi BSP default versions
 
-PREFERRED_VERSION_linux-raspberrypi ??= "4.9.%"
+PREFERRED_VERSION_linux-raspberrypi ??= "4.14.%"
```
agherzan: From the git repository 4.9 still seems to be the default branch. How did we decide on this version's preference?

kraj: This will switch to 4.14 soon'ish, we still have couple of months for sumo release. We can track back to 4.9 if they dont switch default by then, but this will give enough testing for 4.14 from our end

recipes-core/images/rpi-basic-image.bb
kraj: OK thats fine by me

conf/machine/include/rpi-default-versions.inc
betafive: https://github.com/raspberrypi/firmware has been changed to use 4.14.y.

kraj: WIFI/BLE comes up, userland/gstreamer works with WPE video acceleration is good so all my stringent tests seems to pass with 4.14 as of now.

agherzan: Thanks guys.

betafive:
I'll update the first commit message and re-base. Will probably be done on Friday.

agherzan:
@betafive I can do that now. Is that Ok with you?

betafive:
Go for it. The only reason it didn't follow the usual pattern was that I couldn't think of what to put before the ':' as it affects a bbclass and a recipe. Maybe just use the function names?

agherzan:
Merged. Thanks @betafive

PR #202, opened 2018-02-28
Upgrade to 4.14 kernel

kraj:
Firmware and userland updates are needed for 4.14 to work.

### DIFF
--- a/recipes-kernel/linux/linux-raspberrypi_4.14.bb
+++ b/recipes-kernel/linux/linux-raspberrypi_4.14.bb
@@ -1,6 +1,6 @@
 LINUX_VERSION ?= "4.14.22"
 
-SRCREV = "d95cdf3e6a0506e3a728b0c80ae02368779d2768"
+SRCREV = "daa138dae512b374c5be05f492e7c62838359b53"
 SRC_URI = "git://github.com/raspberrypi/linux.git;branch=rpi-4.14.y"
 
 require linux-raspberrypi.inc
--- a/recipes-bsp/common/firmware.inc
+++ b/recipes-bsp/common/firmware.inc
@@ -1,9 +1,11 @@
-RPIFW_DATE ?= "20180209"
-RPIFW_SRC_URI ?= "https://github.com/raspberrypi/firmware/archive/b1a7f4aea6cbd380319c2849ecc5988f9a4dba70.tar.gz"
-RPIFW_S ?= "${WORKDIR}/firmware-b1a7f4aea6cbd380319c2849ecc5988f9a4dba70"
+RPIFW_DATE ?= "20180226"
+SRCREV ?= "469fc794fd3e45dbf30e14b46cb0d6283e399cba"
+RPIFW_SRC_URI ?= "https://github.com/raspberrypi/firmware/archive/${SRCREV}.tar.gz"
+RPIFW_S ?= "${WORKDIR}/firmware-${SRCREV}"
 
 SRC_URI = "${RPIFW_SRC_URI}"
-SRC_URI[md5sum] = "08ed10d92ee064711d73790557267f84"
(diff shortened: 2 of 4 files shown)


kraj:
I rebased it on top of pbarker/next

betafive:
@kraj: Can the firmware & userland commits you've used still support a v4.9.y kernel? I want to keep both options working at least for the next few weeks while things transition.

kraj:
@betafive I have not tested it with 4.9 however, we might have to use stable branch of firmware with 4.9, that probably means another recipe and may be another lockdown of PREFERRED_VERSION if someone choses to stick to 4.9

betafive:
@kraj: The commits I used support both 4.9 and 4.14 kernels - I've boot tested both options. I've not done wifi & bluetooth testing as you have though.

I'd really like to have a point in time where both kernels work so that we can isolate any variation between them if we need to debug. I'm happy if things move to 4.14 only very soon after that though - we don't have the bandwidth to keep both kernels supported for long.

It also gives us one last update of firmware & kernel we can backport to rocko before master goes forward as 4.14 only.

kraj:
I have looked into bootfiles recipe and from quick look it seems the portions we use from the firmware blob, we should be ok with 4.9 as well

recipes-kernel/linux/linux-raspberrypi_4.14.bb
```diff
@@ -1,6 +1,6 @@
 LINUX_VERSION ?= "4.14.22"
 
-SRCREV = "d95cdf3e6a0506e3a728b0c80ae02368779d2768"
```
agherzan: Is now an update and not an added recipe anymore.

betafive:
I'm not too sure on the firmware. I had issues during the 4.4 -> 4.9 transition where 4.4 won't boot with more recent firmware. I'd like to stick to something that has been explicitly boot tested with 4.9 for at least a week or two to give people chance to test things out.

agherzan:
@kraj Do you depend on the revision of firmware you update to in this PR? Would the one Paul has work for you too?

kraj:
I think its better to use the newer one whichever that is. Since rpi org has been pushing 4.14 bugfixes to it. I have tested my set but I guess either would work.

kraj:
I have rebased it again on top of master now

recipes-kernel/linux/linux-raspberrypi_4.14.bb
kraj: yes this is an update on top of what was pushed as new recipe for 4.14

betafive: I think what Andrei means is that the commit message is now confusing.

betafive:
The simplest way to fix this is for me to make some time to boot test this with a 4.9 kernel. I should be able to spare some time on Friday to do that. I'd rather be careful here so we don't disrupt other users unnecessarily.

General review comments:

* Commit message for linux-raspberrypi changes now need to say that it's an update not a new recipe.

* Changes to firmware.inc re-order lines un-necessarily. Can we tidy that up so it's more clear what has changed?

* Userland update looks good.

kraj:
updated

betafive:
Thanks @kraj, I'll test with 4.9 as soon as I get chance and then let's go ahead and merge this if all works.

kraj:
@betafive any updates on this ?

agherzan:
@betafive A friendly ping on this.

betafive:
Was just about to comment on this. I've had a really busy couple of weeks, got some client work to finish off before Friday but then should have some time to look at this again. Apologies for the delay.

betafive:
All looks good. Really sorry this took so long to test, had a few unexpected things come up at work.

agherzan:
Better late... :) Thanks @betafive

PR #205, opened 2018-03-07
bluez5: brcm43438: update device dependency

pabigot:
A previous patch changed the BT serial device to the alias, but did not
update the device dependency resulting in the service failing to start
on boot.

Signed-off-by: Peter A. Bigot <pab@pabigot.com>


agherzan:
ok to test

PR #206, opened 2018-03-07
patches to improve wic support

pabigot:
Many kickstart files used `--on mmcblk` for SD builds, which produces the image but breaks runtime now that wic no longer excludes /boot from the set of partitions added to fstab for mount on boot.  The first commit patches the raspberrypi kickstart file to set the correct device.

The second  commit simply provides a default for `WKS_FILE`, to make it simpler for people to add wic as an image fstype in `local.conf` as with:

    IMAGE_FSTYPES_append = " wic wic.bmap"


betafive:
ok to test

PR #207, opened 2018-03-08
gstreamer1.0-omx: remove unused patches

maxinbjohn:
Remove the following unused patches:
 1. 0002-Don-t-try-to-acquire-buffer-when-src-pad-isn-t-activ.patch
 2. 0003-fix-decoder-flushing.patch

Signed-off-by: Maxin B. John <maxin.john@intel.com>

**- What I did**

**- How I did it**


agherzan:
ok to test

PR #209, opened 2018-03-14
Apply -T flag for xz SD image compression to use all available CPU cores

koolkhel:
I need generated SD image compression so I could transfer the image from my build server in a less amount of time. I found out that compression step takes a long time, although my server is very capable. So I have added -T flag for `xz` call to automatically use all available cores and compress an image in least possible time. It makes sense to update all release branches as well, shouldn't break anything.


koolkhel:
oh, sorry, it should spell "xz -T 0", forgot the 0

PR #210, opened 2018-03-14
Apply -T flag for xz SD image compression to use all available CPU cores

koolkhel:
I need generated SD image compression so I could transfer the image from my build server in a less amount of time. I found out that compression step takes a long time, although my server is very capable. So I have added -T flag for xz call to automatically use all available cores and compress an image in least possible time. It makes sense to update all release branches as well, shouldn't break anything.

Sorry, sending this a second time, my fault.

### DIFF
--- a/classes/sdcard_image-rpi.bbclass
+++ b/classes/sdcard_image-rpi.bbclass
@@ -183,7 +183,7 @@ IMAGE_CMD_rpi-sdimg () {
 		bzip2 -k9 "${SDIMG}"
 		;;
 	"xz")
-		xz -k "${SDIMG}"
+		xz -T 0 -k "${SDIMG}"
 		;;
 	esac
 }


betafive:
How are you enabling xz compression of the SD image? I think there's two ways to do this...

1) Setting `IMAGE_FSTYPES = "rpi-sdimg.xz"`
2) Setting `IMAGE_FSTYPES = "rpi-sdimg"` and `SDIMG_COMPRESSION = "xz"`.

If you use method (1), the conversion command in image_types.bbclass (http://git.openembedded.org/openembedded-core/tree/meta/classes/image_types.bbclass) should already handle multi-threading. We should probably deprecate method (2) and recommend everyone use method (1).

koolkhel:
Indeed, I use
```
WKS_FILES = "sdimage-raspberrypi.wks"
IMAGE_FSTYPES = "rpi-sdimg"
SDIMG_COMPRESSION="xz"
```
I'll switch to method 1, thanks!

agherzan:
@betafive What do you think if we would remove completely SDIMG_COMPRESSION?

betafive:
@agherzan: That would make sense. I think it's completely redundant now.

agherzan:
@koolkhel Would you be able to put up a PR for this?

koolkhel:
@agherzan ok, I'll remove all code paths related to SDIMG_COMPRESSION configuration variable and make a corresponding PR

agherzan:
Thanks, @koolkhel . Please make the changes in the documentation too.

PR #215, opened 2018-03-28
RasperryPi 3B+ support

agherzan:
**- What I did**
Packages updates needed for RasperryPi 3B+ support (and some kernel updates)


agherzan:
@kraj good point on naming. I actually thought to go with the PREFERRED_PROVIDER but I just didn't want to maintain all the other packages that come with the linux-firmware package. Basically just wanted two firmware as they are used by RPI. And this simplified a lot the recipe.

kraj:
@agherzan I think people using extra HATS might need extra firmware no ?

agherzan:
@kraj  Exactly. They will use the ones from linux-firmware for everything but the integrated wifi firmware that will be provided by the raspbian repository.

betafive:
I share @kraj's concern about the name. `linux-firmware-raspbian` would be much clearer.

I don't think we need to worry about packaging everything from the Raspbian firmware repository right now though. If there are extra things needed for various hats we can add them to the recipe later.

Other than that, all looks good.

Kostia-K:
I believe you also need to add `bcm2710-rpi-3-b-plus.dtb` to KERNEL_DEVICETREE in rpi-base.inc

agherzan:
What? They have a new dt? Good catch @Kostia-K. I'll take a look.

agherzan:
@Kostia-K @kraj @betafive Renamed the recipe, added the rpi3b+ dtb, retested, all good.

hhromic:
Hi ! Thanks for the update!
Should the `linux-firmware-raspbian` new package be used also for the `raspberrypi0-wifi.conf` config file aswell? The RPI0W also comes with Wifi/BT onboard.

hhromic:
Also, this new `linux-firmware-raspbian` recipe renders the existing `linux-firmware` recipe obsolete? e.g. should be removed?

https://github.com/agherzan/meta-raspberrypi/tree/master/recipes-kernel/linux-firmware

agherzan:
@hhromic I just left it there for now to not break people's work. If they use linux-firmware they will still get the same blobs. I was tempted to remove them at first.

hhromic:
Good point.
For the `rocko` branch I agree 100% it should be kept.
However, this PR is for the `master` branch, which is expected to be bleeding edge and should not contain deprecated things that in the end turn into more maintenance effort for lazy users that do not read the documentation.
This change should be documented. If you remember from #184, it was not clear that the firmware has to be explicitly included. In this case the documentation should say that "if you want Wifi/BT you must include `linux-firmware-raspbian`.

agherzan:
@hhromic agreed. Pushed changed to address documentation and removal of linux-firmware bits.

kraj:
Looks good

davidweisgerber:
Will this be backported to morty?

agherzan:
There are no plans as far as I am aware of this point. We would not reject patches though.

betafive:
Morty currently defaults to the 4.4 kernel series which won't support RPi 3 B+. The recipe for the 4.9 series kernel and the firmware are also pretty far behind and would need a lot of updates to support that device. From previous testing I'm pretty sure the firmware updates break compatibility with the 4.4 series kernels and so that's probably not a good avenue to go down.

I'd recommend trying out master or the backports for rocko which I've just posted (PR #229).

agherzan:
Good point.

PR #227, opened 2018-04-16
Kernel updates, removal of obsolete config and dependency/task cleanup

betafive:
## Changes

* Apply patch to fix `bitbake linux-raspberrypi -c menuconfig` on distros where the native ncurses differs from the one we build. Discussed on oe-core list here: http://lists.openembedded.org/pipermail/openembedded-core/2018-April/149912.html

* Updates for all our kernel recipes.

* Drop unnecessary dependencies and tasks from a few recipes.

* Switch from RDEPENDS to DEPENDS for recipes which don't create a package.

* Major cleanup of kernel `do_configure_prepend` function. Everything has been reviewed against the kernel config created when running `make bcmrpi_defconfig` and `make bcm2709_defconfig` for `ARCH=arm` and `make bcmrpi3_defconfig` for `ARCH=arm64` in the linux-raspberrypi sources. I've kept all settings which appear to be functional and not horrifically obsolete.

I'm hoping that we can replace everything in `do_configure_prepend` with the use of KERNEL_FEATURES and config fragments as supported by `kernel-yocto.bbclass`. That would allow us to drop the `config_setup` and `kernel_configure_variable` functions as well. This is essentially a first step towards that - we drop obsolete config options so that there's less to convert and so we have a baseline of functional config options to test against.

## Testing
Done: build testing
Done: Boot tested with and without u-boot on RPi 1, 2, 3 and 3-64, plus some basic command line messing around.

I wanted to get this up for review now though so people can comment and in case anyone else has kernel changes in progress. I'd like to get this merged before we branch for the 'sumo' release so we have less obsolete stuff to maintain.

### DIFF
--- a/recipes-bsp/bootfiles/bcm2835-bootfiles.bb
+++ b/recipes-bsp/bootfiles/bcm2835-bootfiles.bb
@@ -9,7 +9,7 @@ include recipes-bsp/common/firmware.inc
 
 INHIBIT_DEFAULT_DEPS = "1"
 
-RDEPENDS_${PN} = "rpi-config"
+DEPENDS = "rpi-config"
 
 COMPATIBLE_MACHINE = "^rpi$"
 
--- a/recipes-bsp/bootfiles/bcm2835-bootfiles.bb
+++ b/recipes-bsp/bootfiles/bcm2835-bootfiles.bb
@@ -3,9 +3,11 @@ LICENSE = "Proprietary"
 
 LIC_FILES_CHKSUM = "file://LICENCE.broadcom;md5=4a4d169737c0786fb9482bb6d30401d1"
 
-inherit deploy
+inherit deploy nopackages
 
 include recipes-bsp/common/firmware.inc
+
+INHIBIT_DEFAULT_DEPS = "1"
 
 RDEPENDS_${PN} = "rpi-config"
 
@@ -32,7 +34,7 @@ do_deploy() {
     touch ${DEPLOYDIR}/${PN}/${PN}-${PV}.stamp
 }
 
-addtask deploy before do_package after do_install
+addtask deploy before do_build after do_install
 do_deploy[dirs] += "${DEPLOYDIR}/${PN}"
 
 PACKAGE_ARCH = "${MACHINE_ARCH}"
--- a/recipes-bsp/u-boot/u-boot_%.bbappend
+++ b/recipes-bsp/u-boot/u-boot_%.bbappend
@@ -4,4 +4,4 @@ SRC_URI_append_rpi = " \
     file://0002-rpi_0_w-Add-configs-consistent-with-RpI3.patch \
 "
 
-RDEPENDS_${PN}_append_rpi = " rpi-u-boot-scr"
+DEPENDS_append_rpi = " rpi-u-boot-scr"
--- a/recipes-kernel/linux/linux-raspberrypi.inc
+++ b/recipes-kernel/linux/linux-raspberrypi.inc
@@ -32,9 +32,6 @@ CMDLINE_append += ' ${@oe.utils.conditional("DISABLE_RPI_BOOT_LOGO", "1", "logo.
 CMDLINE_DEBUG ?= ""
 CMDLINE_append = " ${CMDLINE_DEBUG}"
 
-# Quirk for udev greater or equal 141
-UDEV_GE_141 ?= "1"
-
 # Enable OABI compat for people stuck with obsolete userspace
 ARM_KEEP_OABI ?= "1"
 
@@ -102,11 +99,8 @@ do_configure_prepend() {
     CONF_SED_SCRIPT=""
 
     # oabi / eabi support
-    kernel_configure_variable AEABI y
     if [ "${ARM_KEEP_OABI}" = "1" ] ; then
         kernel_configure_variable OABI_COMPAT y
-    else
-        kernel_configure_variable OABI_COMPAT n
     fi
 
     # Set cmdline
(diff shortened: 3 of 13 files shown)


recipes-bsp/bootfiles/bcm2835-bootfiles.bb
```diff
@@ -16,7 +16,7 @@ do_package_write_rpm[noexec] = "1"
 do_package_write_deb[noexec] = "1"
 do_populate_sysroot[noexec] = "1"
 
-RDEPENDS_${PN} = "rpi-config"
+DEPENDS_${PN} = "rpi-config"
```
kraj: you don't suffix  ```DEPENDS``` with  ```_${PN}```

recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -4,4 +4,4 @@ SRC_URI_append_rpi = " \
     file://0002-rpi_0_w-Add-configs-consistent-with-RpI3.patch \
 "
 
-RDEPENDS_${PN}_append_rpi = " rpi-u-boot-scr"
+DEPENDS_${PN}_append_rpi = " rpi-u-boot-scr"
```
kraj: same as above

recipes-kernel/linux/linux-raspberrypi.inc
```diff
@@ -111,10 +111,6 @@ do_configure_prepend() {
         kernel_configure_variable SQUASHFS y
         kernel_configure_variable UBIFS_FS y
     fi
-    # Activate CONFIG_LEGACY_PTYS
-    kernel_configure_variable LEGACY_PTYS y
-    # this module is built externally via drbd-utils
-    kernel_configure_variable BLK_DEV_DRBD n
```
kraj: I think disabling this is still the right choice, we should document how to enable drbd from meta-networking instead

recipes-kernel/linux/linux-raspberrypi.inc
```diff
@@ -103,9 +103,6 @@ do_configure_prepend() {
         kernel_configure_variable OABI_COMPAT y
     fi
 
-    # Set cmdline
-    kernel_configure_variable CMDLINE "\"${CMDLINE}\""
```
kraj: will I be able to see it in zcat proc/config.gz | grep CMDLINE ?

recipes-bsp/bootfiles/bcm2835-bootfiles.bb
```diff
@@ -7,6 +7,15 @@ inherit deploy
...
 include recipes-bsp/common/firmware.inc
 
+INHIBIT_DEFAULT_DEPS = "1"
+do_package[noexec] = "1"
+do_package_qa[noexec] = "1"
+do_packagedata[noexec] = "1"
+do_package_write_ipk[noexec] = "1"
+do_package_write_rpm[noexec] = "1"
+do_package_write_deb[noexec] = "1"
+do_populate_sysroot[noexec] = "1"
```
kraj: can we see if ```deltask``` is the right option here ?

recipes-kernel/linux/linux-raspberrypi.inc
```diff
@@ -102,51 +99,15 @@ do_configure_prepend() {
     CONF_SED_SCRIPT=""
 
     # oabi / eabi support
-    kernel_configure_variable AEABI y
     if [ "${ARM_KEEP_OABI}" = "1" ] ; then
         kernel_configure_variable OABI_COMPAT y
-    else
-        kernel_configure_variable OABI_COMPAT n
     fi
```
kraj: We should document what kernels this change will work on. I am fine with change to limit itself to newer kernels.

recipes-kernel/linux/linux-raspberrypi.inc
```diff
@@ -109,20 +109,6 @@ do_configure_prepend() {
...
-        kernel_configure_variable INET y
-        kernel_configure_variable IP_PNP y
-        kernel_configure_variable USB_GADGET y
-        kernel_configure_variable USB_GADGET_SELECTED y
-        kernel_configure_variable USB_ETH y
-        kernel_configure_variable NFS_FS y
-        kernel_configure_variable ROOT_NFS y
-        kernel_configure_variable ROOT_NFS y
-        kernel_configure_variable CMDLINE "\"${CMDLINE_NFSROOT_USB}\""
-    fi
```
kraj: agreed, although such a feature to set cmdline based on kernel options would be nice to have.

recipes-kernel/linux/linux-raspberrypi.inc
betafive: Not sure I understand your comment here. This is just removing the assignment of `OABI_COMPAT=n` when `ARM_KEEP_OABI` is not enabled since the default is already `n` in the kernel.

Long-term we should stop enabling this by default but I don't want to make such a potentially breaking change right now.

recipes-kernel/linux/linux-raspberrypi.inc
betafive: No, but I think the CONFIG_CMDLINE variable in /proc/config.gz is meaningless anyway. You'd want to look at `/proc/cmdline` for the command line actually used at runtime.

recipes-kernel/linux/linux-raspberrypi.inc
betafive: We already have logic to populate `cmdline.txt` in the boot partition based on the `CMDLINE` variable, so this just needs documenting.

betafive:
@kraj: Thanks for the quick feedback! Will address those points and update.

recipes-bsp/bootfiles/bcm2835-bootfiles.bb
betafive: Reading https://www.yoctoproject.org/docs/2.4.2/bitbake-user-manual/bitbake-user-manual.html#deleting-a-task, I'd say `noexec` is the safer option as it ensures we don't accidentally lose any dependency tasks.

recipes-kernel/linux/linux-raspberrypi.inc
kraj: I meant to say that if we are providing a knob to be able to use OABI that means we are supporting this ABI which seems to be not right. We should get rid of this setting completely, in long run we should be switch to yocto kernel tooling and drop the custom ways

recipes-bsp/bootfiles/bcm2835-bootfiles.bb
kraj: yes however, deleting means they dont show up in queue at all which is even better from build point of view, so yes asjust the dependencies as required and delete the tasks

recipes-kernel/linux/linux-raspberrypi.inc
kraj: this lets us use meta-networking out of box, if there are no users of drbd who do not use meta-networking as well then your change makes sense, but thats a smaller set than the other

recipes-kernel/linux/linux-raspberrypi.inc
kraj: ok

recipes-kernel/linux/linux-raspberrypi.inc
betafive: Ok. I'm trying to avoid us carrying any config changes we don't have a good reason for. Supporting the recipes in meta-networking out-of-the-box is probably worthwhile though. I'll add this back in but keep CONFIG_LEGACY_PTYS disabled.

recipes-bsp/bootfiles/bcm2835-bootfiles.bb
betafive: Fixed.

recipes-bsp/u-boot/u-boot_%.bbappend
betafive: Fixed.

recipes-kernel/linux/linux-raspberrypi.inc
betafive: Fixed.

recipes-kernel/linux/linux-raspberrypi.inc
betafive: I'm planning to leave my change as-is. That ok?

recipes-bsp/bootfiles/bcm2835-bootfiles.bb
betafive: Fixed.

recipes-kernel/linux/linux-raspberrypi.inc
betafive: Fixed. I've added a final commit to drop OABI compatibility so it's easy to back out if that becomes necessary.

recipes-kernel/linux/linux-raspberrypi.inc
betafive: Fail. drbd in meta-networking doesn't build against Linux v4.14.y. The version in meta-networking is v9.0.8, compatibility with this kernel version was added in v9.0.10 according to drbd release notes (https://github.com/LINBIT/drbd-9.0/blob/master/ChangeLog).

So I think we should go back to leaving BLK_DEV_DRBD enabled.

betafive:
Changes just pushed:

* Rebased on latest master.
* Switched to using `nopackages.bbclass` instead of manually calling `deltask` or setting `noexec`. This should be better going forward as the nopackages class is maintained in oe-core.
* Dropped the forced setting of `CONFIG_BLK_DEV_DRBD=n` again. We should leave this enabled as a module by default as the drbd recipe in meta-networking doesn't build against Linux 4.14.y.

Boot tested with and without u-boot on RPi 1, 2, 3 and 3-64.

PR #229, opened 2018-04-25
Rocko Backports

betafive:
Apologies these are so late, things have been busy!

## Changes

Changes to the kernel and firmware are backported along with a couple of other minor fixes.

I've left out the following which could be candidates for backporting. These aren't really things I use heavily though so advice and testing from others would probably be needed to help here:
* bluez5
* userland
* linux-firmware-raspbian

## Testing

I've booted tested on all RPi variants I have but that doesn't include an RPi 3 B+.

Testing from anyone with access to an RPi 3 B+ would be much appreciated! Otherwise you'll all have to wait for me to order one some time next week. I wanted to make sure people see some progress here though as I know I'm a bit behind on backports.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,10 +10,11 @@ Yocto BSP layer for the Raspberry Pi boards - <http://www.raspberrypi.org/>.
 ## Quick links
 
 * Git repository web frontend:
-  <http://git.yoctoproject.org/cgit/cgit.cgi/meta-raspberrypi/>
+  <https://github.com/agherzan/meta-raspberrypi>
 * Mailing list (yocto mailing list): <yocto@yoctoproject.org>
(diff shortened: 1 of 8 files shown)


davidweisgerber:
Hi, for me it boots to raspberry pi 3b+ using my custom image (with all sort of Qt 5 stuff and eglfs)
Bluetooth is working but WiFi does not.

davidweisgerber:
I found the firmware for the WiFi seems to be working. I will continue testing.
Found a bug: reboot is not working

davidweisgerber:
Reboot problem was due to upgrading the build system. For me now it works now

betafive:
Thanks for the testing!

PR #233, opened 2018-05-01
simplify use of usb gadget mode

pabigot:
When initially provisioning a headless device it's nice to be able to connect it to a computer using USB networking.  This requires using a USB driver that understands peripheral (gadget) mode.  These patches add the required dt overlay and provide a short-hand to add it to config.txt.

### DIFF
--- a/docs/extra-build-config.md
+++ b/docs/extra-build-config.md
@@ -221,13 +221,21 @@ needs a fixed core frequency and enable_uart wil set it to the minimum. Certain
 operations - 60fps h264 decode, high quality deinterlace - which aren't
 performed on the ARM may be affected, and we wouldn't want to do that to users
 who don't want to use the serial port. Users who want serial console support on
-RaspberryPi3 will have to explicitely set in local.conf:
+RaspberryPi3 will have to explicitly set in local.conf:
 
     ENABLE_UART = "1"
 
 Ref.:
 * <https://github.com/raspberrypi/firmware/issues/553>
 * <https://github.com/RPi-Distro/repo/issues/22>
+
+## Enable USB Peripheral (Gadget) support
+
+The standard USB driver only supports host mode operations.  Users who
+want to use gadget modules like g_ether should set the following in
+local.conf:
+
+    ENABLE_DWC2_PERIPHERAL = "1"
 
 ## Manual additions to config.txt
 
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -24,6 +24,7 @@ KERNEL_DEVICETREE ?= " \
     bcm2708-rpi-cm.dtb \
     bcm2710-rpi-cm3.dtb \
     \
+    overlays/dwc2.dtbo \
     overlays/hifiberry-amp.dtbo \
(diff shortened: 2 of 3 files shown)


docs/extra-build-config.md
```diff
@@ -221,14 +221,22 @@ needs a fixed core frequency and enable_uart wil set it to the minimum. Certain
...
 * <https://github.com/raspberrypi/firmware/issues/553>
 * <https://github.com/RPi-Distro/repo/issues/22>
 
+## Enable USB Peripheral (Gadget) support
+
+The standard USB driver only supports host mode operations.  Users who
+want to use gadget modules like g_ether should set the following in
+local.conf:
+
+    ENABLE_DWC2_PERIPHERAL = "1"
```
kraj: Can we call it something that conveys the feature easily something like like OTG

secondly, should we add usbinit to image automatically and check if usbgadget is  in DISTRO/MACHINE_FEATURES

pabigot: Re `ENABLE_OTG`: We're on the periphery of my understanding here, so I used a name that reflects the specifics of what this actually does.  It may be better to reserve `ENABLE_OTG` for something that adds `dtoverlay=dwc` with or without `,dr_mode=otg` to make the module default explicit.   I've only tested `dr_mode=peripheral`; I don't want to imply that if an OTG cable is used it'll work as a host.

I suppose I'd better try `dr_mode=otg` and if what I need works switch to that; I agree the name would be more clear.

Re usbinit. ...maybe?  I wasn't aware of that, and haven't used it; I just put a `g_ether.conf` file in `/etc/modules-load.d/`.  Reviewing the `usb-gether` script my concern is with the ifup/ifdown  commands.  I'm using connman, so those aren't necessary.  Also I'm unsure how to do the check and where to put the conditions.  @kraj Can you point me to an example BSP layer that does what you're proposing?

kraj: look into OE-Core and poky uses it, I am not sure if any specific BSP layer I know of is using it
but its supposed to be

pabigot: I did find `usbinit` in poky, but no references to it related to usbgadget in features anywhere.  It's just a recipe that's available.  So absent further information I think that part is out of scope

pabigot:
After further research and testing I'm standing by the version as submitted:

* Most references to gadget support on RPi explicitly select `dtoverlay=dwc2,dr_mode=peripheral`.  IMO using `ENABLE_OTG` would be confusing as people might expect `dr_mode=otg`, which is the module default.
* When I use `dr_mode=otg` or leave it as the default with g_ether enabled connecting the RPi to another host causes that host to experience USB errors (hosts being Ubuntu 18.04/Linux 4.15.0-20-generic, two different machines).  On one machine the errors take down the entire USB infrastructure.  I have not observed this when the module is loaded in peripheral mode.

kraj:
thanks for the additional information. I am fine with otg vs peripheral change lets keep it as you have proposed. Although I still think that we can somehow use ```usbinit``` automatically when it is enabling dwc this will bringup usb0 automatically on boot, I think that would be ideal. May be append it to ```MACHINE_EXTRA_RRECOMMENDS```

PR #238, opened 2018-05-07
Firmware/Kernel updates then rework bluetooth to fix licensing and address issues

pabigot:
Tested on 3B+ and 0-wifi.

firmware updated because there's a 3B+ patch in there that might affect identifying the pins for hardware flow control in the bluetooth uart (which is now supported).  Kernel updated to match the firmware, plus one more release to reduce the amount of lan78xx noise.  These may not be necessary, but that's how I tested everything.

A couple patches to the existing bluez recipe and systemd script, then the patches that break out the material that needs its own license and supports using the standard Raspbian approach to configuring the bluetooth device.


recipes-connectivity/bluez5/bluez5_%.bbappend
```diff
@@ -52,4 +52,4 @@ FILES_${PN}_append_raspberrypi0-wifi = " ${BCM_BT_FIRMWARE}"
 
 SYSTEMD_SERVICE_${PN}_append_raspberrypi0-wifi = " ${BCM_BT_SERVICE}"
 
-RDEPENDS_${PN}_append_raspebrrypi0-wifi = " udev-rules-rpi"
+RDEPENDS_${PN}_append_raspberrypi0-wifi = " udev-rules-rpi"
```
kraj: I know this is not related to the given change proposed. But I question pi0 specific RDEPENDS, I would think it
should be coming via MACHINE_EXTRA_RRECCOMMENDS or some such.

recipes-connectivity/bluez5/bluez5/brcm43438.service
```diff
@@ -2,10 +2,12 @@
 Description=Broadcom BCM43438 bluetooth HCI
 ConditionPathIsDirectory=/proc/device-tree/soc/gpio@7e200000/bt_pins
 Before=bluetooth.service
+Wants=dev-serial1.device
```
kraj: perhaps this should be
```
BindsTo=
```

recipes-connectivity/bluez5/bluez5_%.bbappend
pabigot: I'm not in love with that mostly because it replicates a series of target-specific capabilities.  However, without introducing some new OVERRIDES tag that identifies the subset of raspberry pi devices that have built-in bluetooth devices I didn't see a nice way to clean it up, other than to move the list of dependencies out to a variable which is where I left it.

pabigot:
BTW: When merged I think this should go to sumo as well, primarily to fix the licensing issues.

pabigot:
Rebased, corrected summary in one commit message, and switched to BindsTo.

kraj:
LGTM now

recipes-connectivity/bluez5/bluez5/brcm43438.service
agherzan: @kraj Why do you recommend `BindsTo`?

recipes-connectivity/bluez5/bluez5_%.bbappend
agherzan: @kraj you have a point there but the issue is the fact that bluez brings the systemd service which depends on the udev rules. So this makes bluez package dependent on the udev rules. That's why we have that RDEPENDS, at runtime, the systemd service needs the udev rules. Probably the clean way forward would be to get the service out in another recipe and make that depend on udev rules but I wonder if that would not confuse people more.

pabigot: Actually with the rework it's pi-bluetooth that depends on the device name and hence on the udev script that provides the alias.  So that should be updated.

pabigot: Or maybe not: pi-bluetooth won't work if there isn't a /dev/serial1 but it'll run without it, just erroring out.  Maybe RRECOMMENDS is a better approach, though I don't want to tackle it myself/now.

recipes-connectivity/bluez5/bluez5/brcm43438.service
kraj: to express correct deps

https://www.freedesktop.org/software/systemd/man/systemd.unit.html#BindsTo=

agherzan: Sure, but how does it help? The bind of usually makes a difference when you want to stop a service if it's dependency stops. But there is nothing to stop as this is just a script. I'm not sure how this applies in this case.

pabigot: I went with it for primarily because it really is a strong dependency: if the device goes away (which might happen), than the HCI configuration isn't valid anymore.  As I read `BindsTo` it represents this and would reflect it in the systemd state.  I don't think `Wants` wouldn't even stop the unit from running if the device wasn't there.

recipes-connectivity/bluez5/bluez5_%.bbappend
agherzan: Agreed.

PR #242, opened 2018-05-09
rebase master pr #238 onto sumo

pabigot:
Suggest this be done as otherwise the sumo branch has questionable licensing for bluez5.


agherzan:
Thanks, @pabigot - merged to sumo.

PR #245, opened 2018-05-11
master: linux-firmware: restore and merge bluez-firmware

pabigot:
As an exercise I came up with this to fix the problems with installing the roll-up linux-firmware package.  It does end up removing both linux-firmware-raspbian and bluez-firmware, rolling the content they provided into linux-firmware.  It appears to resolve the problem in my environment.

I've chosen to not use overrides to conditionalize it on machine as had been done in the previous version; IMO this is a BSP layer and there's no need to make it a no-op if the target isn't a Raspberry Pi device.  To discuss.

If this is the way people want to go there may be need for RPROVIDES settings to allow these to supersede the removed packages if anybody's got package repositories.  If so I'd need guidance on how to do that.


recipes-kernel/linux-firmware/linux-firmware_%.bbappend
```diff
@@ -0,0 +1,70 @@
...
+
+SRCREV_raspbian-nf = "86e88fbf0345da49555d0ec34c80b4fbae7d0cd3"
+SRCREV_raspbian-bluez = "e28cd7ee8615de33aa7ec2b41d556af61a4a2707"
+SRCREV_FORMAT = "git"
+
+do_install_append() {
+    install -d ${D}${nonarch_base_libdir}/firmware/brcm/
+
+    # Replace outdated upstream files with fresh from raspbian firmware-nonfree
+    for fw in brcmfmac43430-sdio.bin brcmfmac43455-sdio.bin ; do
```
hhromic: Don't forget the `.txt` NVRAM files :)
Maybe better is to use globs, e.g. `brcmfmac43430-sdio*` and `brcmfmac43455-sdio` ?

recipes-kernel/linux-firmware/linux-firmware_%.bbappend
```diff
@@ -0,0 +1,70 @@
...
+    ${PN}-cypress-license \
+    ${PN}-bcm43430a1-hcd \
+    ${PN}-bcm4345c0-hcd \
+"
+
+# For additional Broadcom
+LICENSE_${PN}-bcm43455 = "Firmware-broadcom_bcm43xx"
+
+FILES_${PN}-bcm43455 = " \
+  ${nonarch_base_libdir}/firmware/brcm/brcmfmac43455-sdio.bin \
```
hhromic: Same here about the `.txt` NVRAM files

recipes-kernel/linux-firmware/linux-firmware_%.bbappend
hhromic: For reference, see PR #194

pabigot: Those were not in `linux-firmware-raspbian`, and they're not packaged in the OE-core `linux-firmware-bcm43430` so I didn't think they were necessary.  But I confirm that the kernel driver will look for them, so I'll make sure they get packaged (they aren't present in the upstream `linux-firmware`, only in the Raspbian fork).  Thanks.

pabigot:
Updated to package ancillary firmware files.

betafive:
> I've chosen to not use overrides to conditionalize it on machine as had been done in the previous version; IMO this is a BSP layer and there's no need to make it a no-op if the target isn't a Raspberry Pi device. To discuss.

I have about 4 BSP layers enabled so I can simply change `MACHINE` and rebuild for the different devices my distro supports. So yes, this does need to be a no-op if `MACHINE` is not one of the ones defined in this layer.

recipes-kernel/linux-firmware/linux-firmware_%.bbappend
```diff
@@ -0,0 +1,75 @@
...
+        install -m 0644 ${WORKDIR}/raspbian-bluez/broadcom/${fw} ${D}${nonarch_base_libdir}/firmware/brcm/
+    done
+}
+
+PACKAGES =+ "\
+    ${PN}-bcm43455 \
+    ${PN}-cypress-license \
+    ${PN}-bcm43430a1-hcd \
+    ${PN}-bcm4345c0-hcd \
+"
```
hhromic: Hi again, shouldn't here be also a `${PN}-bcm43430` entry aswell?

pabigot: No, there's one in the base package from OE-core.  We're just adding a file to it (and changing the file that it packages).

hhromic: Ah you are right, forgot to consider the upstream recipe. Sorry!

recipes-kernel/linux-firmware/linux-firmware_%.bbappend
```diff
@@ -0,0 +1,75 @@
...
+
+# For additional Broadcom
+LICENSE_${PN}-bcm43455 = "Firmware-broadcom_bcm43xx"
+
+FILES_${PN}-bcm43430 += " \
```
hhromic: I remember that in the past version of the recipe, where I provided a PR to include the NVRAM file, I had to use `_append` instead of `+=`, otherwise the build would't include the file. I don't know exactly why.

Check here: https://github.com/hhromic/meta-raspberrypi/blob/fc964659c46b50b69551eadd834d21c81777b00b/recipes-kernel/linux-firmware/linux-firmware_%25.bbappend#L24

pabigot: That was due to the presence of the machine override on the LHS.  Since those are coming back, this will be switched to `_append`.

hhromic: Ah right, the `_rpi` override. I remember now. Thanks for your patience :)

pabigot: No problem; I'm learning as I go....

gizero:
>> I've chosen to not use overrides to conditionalize it on machine as had been done in the previous version; IMO this is a BSP layer and there's no need to make it a no-op if the target isn't a Raspberry Pi device. To discuss.

>I have about 4 BSP layers enabled so I can simply change MACHINE and rebuild for the different devices my distro supports. So yes, this does need to be a no-op if MACHINE is not one of the ones defined in this layer.

Same here! machine overrides are considered best practices for every BSP layer: we definitely need to keep them IMO.

pabigot:
@gizero @betafive fair enough.  Perhaps a documentation bug should be filed against [bsp-guide](https://www.yoctoproject.org/docs/latest/bsp-guide/bsp-guide.html#customizing-a-recipe-for-a-bsp) as it doesn't mention this, and it's hard to search arbitrary BSP layers to find examples to follow.

I'm making progress on this, but like everything else it's not as easy as it appears: things work with `PACKAGES =+` but not with `PACKAGES_append`, which may be due to some special processing done by the base recipe.  I'll keep trying....

gizero:
>Perhaps a documentation bug should be filed against bsp-guide as it doesn't mention this, and it's hard to search arbitrary BSP layers to find examples to follow.

Good advice @pabigot ! Most of what I learned about "fair" BSP-layers comes from the very good series of talks titled "Designing a distro from scratch using OpenEmbedded" by @koenkooi  

>I'm making progress on this, but like everything else it's not as easy as it appears: things work with PACKAGES =+ but not with PACKAGES_append, which may be due to some special processing done by the base recipe. I'll keep trying....

I believe I know what you are dealing with... variable redifinition/ovverides can quickly get a mess... Moreover, the offending way of using them can be buried anywhere, possibly in your layer's dependencies (especially in deeper layer stacks). Can't say anything but suggesting to keep an eye on any source of configuration change in `bitbake -e`

docs/layer-contents.md
```diff
@@ -23,12 +23,9 @@ For other uses it's recommended to base images on `core-image-minimal` or
...
-for these boards. This is because the upstream `linux-firmware` doesn't support
-or has outdated files for the blobs we need. The recipe
-`linux-firmware-raspbian` is based on a fork of `linux-firmware` which
-includes everything we need in order to fully support the WiFi hardware. All
-machines define `MACHINE_EXTRA_RRECOMMENDS` accordingly.
+## WiFi and Bluetooth Firmware
+
+Be aware that the WiFi and Bluetooth firmware for the supported boards
+is not available in stock `linux-firmware`.  These are added from
+Raspbian repositories.  All machines define `MACHINE_EXTRA_RRECOMMENDS`
```
agherzan: NIT: Extra space before `All`.
As well I would reword `stock` by referencing the upstream provider of that recipe, which is poky.

gizero:
@pabigot They're probably crying because of #246

recipes-kernel/linux-firmware/linux-firmware_%.bbappend
```diff
@@ -0,0 +1,77 @@
...
+"
+
+SRCREV_raspbian-nf = "86e88fbf0345da49555d0ec34c80b4fbae7d0cd3"
+SRCREV_raspbian-bluez = "e28cd7ee8615de33aa7ec2b41d556af61a4a2707"
+SRCREV_FORMAT_rpi = "default+raspbian-nf+raspbian-bluez"
+
+do_install_append_rpi() {
+    install -d ${D}${nonarch_base_libdir}/firmware/brcm/
+
+    # Replace outdated upstream files with fresh from raspbian
```
agherzan: Something needs rephrasing here. Maybe have something like `Replace outdated upstream files with the ones from raspbian`.
Extra space before `Raspbian` too.

agherzan:
True:

```
ERROR: No recipes available for:
  /var/lib/jenkins/workspace/meta-raspberrypi-generic/poky/../meta-raspberrypi/recipes-multimedia/gstreamer/gstreamer1.0-omx_1.12%.bbappend
Parsing of 2076 .bb files complete (0 cached, 2076 parsed). 2922 targets, 101 skipped, 0 masked, 0 errors.
```

I'll try to see if it's a trivial update for gst packages.

docs/layer-contents.md
pabigot: I'm of the "Two spaces after period" generation.  I can change it, but can't promise it won't recur.

pabigot:
@gizero @agherzan re crying--thanks.  Couldn't figure out why it was OK last night, and not this morning.  I don't use graphics, so hadn't hit this issue.

docs/layer-contents.md
agherzan: Got it. I can't promise that I won't complain :) . Thanks.

agherzan:
@pabigot are you using current poky master?

pabigot:
@agherzan no, I hadn't updated since yesterday afternoon.  I see it now, and am going back to a base version that works.

pabigot:
Fixed spacing and clarified references to upstream sources.

agherzan:
Thanks @pabigot . Will merge as soon as we fix the builds.

agherzan:
@pabigot  Jenkins should be happy now. Can you please rebase the PR?

pabigot:
@agherzan pushed and checked on 3B+ and 0Wifi.

agherzan:
Thanks @pabigot . Merged to master.

PR #249, opened 2018-05-12
sumo: linux-firmware: restore and merge bluez-firmware

pabigot:
Rebased pull request #245 onto sumo.  The master version should be verified for some period before this is merged, but I think this change should go into sumo ASAP to reduce the amount of package feed churn in a release branch.


agherzan:
Thanks @pabigot

PR #250, opened 2018-05-12
gstreamer1.0-plugins-bad_%.bbappend: Drop --disable-dispmanx

kraj:
1.14 does not support this option
Fixes
QA Issue: gstreamer1.0-plugins-bad: configure was passed unrecognised options: --disable-dispmanx [unknown-configure-option]

**- What I did**

**- How I did it**

### DIFF
--- a/recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
+++ b/recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
@@ -9,4 +9,4 @@ PACKAGECONFIG_GL_rpi = "egl gles2"
 
 PACKAGECONFIG_append_rpi = " hls libmms faad"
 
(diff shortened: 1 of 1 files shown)


shr-project:
This doesn't look correct.

Why do you keep --enable-dismanx option? I think whole dispmanx support moved from bad to base gst plugins.

See https://cgit.freedesktop.org/gstreamer/gst-plugins-bad/commit/?h=1.14&id=769a21d0bb310906b880c07da0e1e2376e11c187

kraj:
Yes that’s a follow up but needs more work and this commit was to disable the QA errors in interim but i think removing the whole config would have been better

shr-project:
Just moving whole PACKAGECONFIG as it was from gstreamer1.0-plugins-bad_%.bbappend to gstreamer1.0-plugins-base_%.bbappend should work.

Issue #254, opened 2018-05-14
need support for raspberrypi3b+ 64


pabigot:
`rpi-config` has an override for `raspberrypi3-64` that specifies the device tree in `cmdline.txt`.  It uses `bcm2710-rpi-3-b.dtb` but for 3B+ it must use `bcm2710-rpi-3-b-plus.dtb` in order to get the right MAC address for lan78xx.

There doesn't appear to be an existing distinction between 3B and 3B+ in meta-raspberrypi.  In the current kernel they have different device trees for both arm and arm64.

pabigot:
This may not be necessary right now, as it appears that simply removing the explicit specfication of device tree in `config.txt` works.

Issue #251, opened 2018-05-13
 RPI 3B+ network not working 


GreSuh:
Build image core-image-full-cmdline for Raspberry3-64.
Image starts without errors but dhcp doesn't work. 

With tcpdump enabled and eth0 in promiscous mode DHCP works.
Seems linke some initialization of eth0 is missing.

**Steps to reproduce the issue:**
Build image  core-image-full-cmdline
openembedded-core
meta-poky
meta-oe
meta-python
meta-perl
meta-networking
meta-security
meta-raspberrypi_priv https://github.com/GreSuh/meta-raspberrypi

but same problem with latest https://git.yoctoproject.org/cgit/cgit.cgi/meta-raspberrypi/ layer

**Describe the results you received:**
no IP Address

**Describe the results you expected:**
System with IP-Address

**Additional information you deem important (e.g. issue happens only occasionally):**

**Additional details (revisions used, host distro, etc.):**
sumo
DISTRO 
poky 
DL_DIR 
/build/downloads 
IMAGE_FSTYPES 
rpi-sdimg tar.bz2 ext4 
IMAGE_INSTALL_append 
kernel-modules  
PACKAGE_CLASSES 
package_rpm 
SSTATE_DIR 
/build/sstate-cache 
PREFERRED_PROVIDER_virtual/kernel
linux-raspberrypi-dev 
RPI_USE_U_BOOT
1

GreSuh:
same problem with default kernel

pabigot:
I suspect a problem with the startup infrastructure.  When I boot the current sumo eth0 works fine, but the MAC address is not the one assigned to the device.  `/proc/cpuinfo` on other machines provides a serial number from which network addresses are derived; this is not present in the raspberrypi-64 build.  Interestingly the wlan0 mac address is correct.

Check whether the mac address you get on boot matches what you expect in your DHCP server.

pabigot:
This is probably #253.

GreSuh:
> This is probably #253.
In my setup i updated the machine definition to match raspberry3 device list.

agherzan:
I hit the same issue lately. It should get fixed by @pabigot 's PR.

agherzan:
@GreSuh this issue should be fixed now in sumo and master branches. Please give it a try.

GreSuh:
Thanks for todays changes on sumo. Network seems to work now. :)

agherzan:
@GreSuh we owe @pabigot a :beer:

PR #255, opened 2018-05-14
sumo: fix firmware support on aarch64

### LINKED ISSUE #253: raspberrypi3-64 not handled completely in MACHINEOVERRIDES
raspberrypi3-64.conf only sets `raspberrypi3` as an override.  It also needs `raspberrypi3-64` or the boot files don't have the proper `config.txt` settings including device tree.

pabigot:
This resolves #253 on sumo and has been tested.

It is not sufficient by itself to address #251 as #254 must also be resolved (I made this change by hand for testing).


pabigot:
Update per review in #256 and add fix satisfying immediate need for #254.

PR #271, opened 2018-06-21
linux-firmware: update Cypress license checksum

hhromic:
Fixes #268

PR #272, opened 2018-06-22
linux-firmware update LIC_FILES_CHKSUM for d114732 revision from oe-core

shr-project:
* unfortunately the LICENCE.cypress was updated in last upgrade:
  http://git.openembedded.org/openembedded-core/commit/?id=6be8744d1b8ee35eb47acd517cfa29b2a7f455d5
  ba51e86 Update Cypress license termination clause
  https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/commit/LICENCE.cypress?id=ba51e861f4444f51e7e83f778575a8146dc514d0

* it's unfortunate, because now meta-raspberrypi/master is compatible only with oe-core master after 6be8744d1b8ee35eb47acd517cfa29b2a7f455d5
  it would be nice to upstream some of these changes to linux-firmware recipes and leave only the rpi specific packaging here

### DIFF
--- a/recipes-kernel/linux-firmware/linux-firmware_%.bbappend
+++ b/recipes-kernel/linux-firmware/linux-firmware_%.bbappend
@@ -8,7 +8,7 @@ LICENSE_append_rpi = "\
 "
 
 LIC_FILES_CHKSUM_append_rpi = "\
-    file://LICENCE.cypress;md5=cbc5f665d04f741f1e006d2096236ba7 \
+    file://LICENCE.cypress;md5=48cd9436c763bf873961f9ed7b5c147b \
 "
 NO_GENERIC_LICENSE[Firmware-cypress] = "LICENCE.cypress"
 


hhromic:
I submitted a PR yesterday #271  addressing the same issue.
I agree with you that this creates a strong dependency on the recent revisions.

kraj:
seems same as https://github.com/agherzan/meta-raspberrypi/pull/271 although commit msg is better

hhromic:
can't argue, indeed the commit msg is better :)

gizero:
The conventional colon after the recipe name in the commit msg would make aesthetes happy as well! ;-)

agherzan:
My queue had https://github.com/agherzan/meta-raspberrypi/pull/271/files as the first PR to look into and found this afterwards. Sadly, even though this commit log is nicer, the other PR got it. What we can do is add a revert to this PR and merge it. Obviously, the laternative is to just close this and blame Andrei. What do you guys think? @shr-project @hhromic ?

hhromic:
My bad actually, I should have closed the other PR after this one was submitted.
The good news is that Github keeps track of this and the other PR in the issue, so if we ever need to come back to this, we have means to not loose track. I will clarify in the issue and the merged PR to check the explanations in this PR for archival purposes.
That said, my opinion is that it is best to not revert just for the sake of commit log clarification.

shr-project:
@hhromic @agherzan The same oe-core change was backported to sumo today, so this or @hhromic's version of this should be backported to meta-raspberrypi/sumo as well.

shr-project:
@agherzan  @hhromic  I've created new PR for that https://github.com/agherzan/meta-raspberrypi/pull/291

hhromic:
just read these messages now.. thanks!

Issue #278, opened 2018-07-04
linux-raspberrypi: devicetree build failing with recent Poky commit


hhromic:
Hello, a recent commit in Poky broke the build of `linux-raspberrypi`:
```
make[3]: *** No rule to make target 'arch/arm/boot/dts/dwc2.dtbo'.  Stop.
arch/arm/Makefile:345: recipe for target 'dwc2.dtbo' failed
make[2]: *** [dwc2.dtbo] Error 2
Makefile:146: recipe for target 'sub-make' failed
make[1]: *** [sub-make] Error 2
Makefile:24: recipe for target '__sub-make' failed
make: *** [__sub-make] Error 2
ERROR: oe_runmake failed
WARNING: exit code 1 from a shell command.
```
The [commit in Poky](https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/?id=396e269c5c0e285be9d5ab40b521dda21bd72e19):
```
commit 396e269c5c0e285be9d5ab40b521dda21bd72e19
Author: Paulo Neves <ptsneves@gmail.com>
Date:   Mon Jul 2 17:03:44 2018 +0200

    kernel-devicetree: Corrected normalize_dtb

    The normalize_dtb function was buggy because
    it only converted from .dts suffix to .dtb
    suffix if the user passed a full source path to
    KERNEL_DEVICETREE containing the /dts/ path.
    The problem is that if the user did that there
    would be a warning.

    On the othet hand if user just set the variable
    KERNEL_DEVICETREE="file.dts" the bbclass translation
    to the respective .dtb target did not occur and
    make would fail saying it has no rule to make target
    file.dts

    This patch decouples the logic of having /dts/ in the
    path from the target translation.

    (From OE-Core rev: 2e7f3b2b9318d1e5395ad58131eafb873f614326)

    Signed-off-by: Paulo Neves <ptsneves@gmail.com>
    Signed-off-by: Richard Purdie <richard.purdie@linuxfoundation.org>
```
I will investigate a PR but if someone knows how to fix it quickly, is welcome :)

hhromic:
Looks like the function in `meta/classes/kernel-devicetree.bbclas` from Poky:
```
normalize_dtb () {
        DTB="$1"
        if echo ${DTB} | grep -q '/dts/'; then
                bbwarn "${DTB} contains the full path to the the dts file, but only the dtb name should be used."
        fi
[...]

PR #279, opened 2018-07-04
Update kernel/userland/firmware to latest as of 20180702

kraj:


hhromic:
the checks are failing because of #278  :(

PR #303, opened 2018-08-15
Modify warning message for deprecated rpi-hwup-target

stevep-korg:
…ge rpi-hwup-image, to recommend using core-image-base.

### DIFF
--- a/recipes-core/images/rpi-hwup-image.bb
+++ b/recipes-core/images/rpi-hwup-image.bb
@@ -7,5 +7,5 @@ IMAGE_INSTALL += " \
 	"
 
 do_image_prepend() {
-    bb.warn("The image 'rpi-hwup-image' is deprecated, please use 'core-image-minimal' instead")
+    bb.warn("The image 'rpi-hwup-image' is deprecated, please use 'core-image-base' instead")
 }


stevep-korg:
OK, I think my final edit to the commit message is acceptable.

Sorry for the first 2 failed attempts.

kraj:
top line in commit message seems redundant, please delete it. You message should be
```text
rpi-hwup-image.bb: Modified the warning message in do_image_prepend()

When bitbaking image rpi-hwup-image, a warning message is generated
to inform he user that the image is deprecated.  The warning message
contains a suggestion of a non-deprecated image to use instead.

Previous to this commit, core-image-minimal was the suggested alternate
image.  This was a subpar suggestion, because rpi-hwup-image packages
the kernel modules in the image, but core-image-minimal does not.  This
can lead to confusion for the user, who may not immediately realize that
their kernel modules are no longer in the image.  With this commit, the
warning message now suggests core-image-base, which packages the kernel
modules in its image.

This change was suggested by Khem Raj on the Yocto project mailing list.

Signed-off-by: Steve Pavao <stevep@korgrd.com>
```

agherzan:
What @kraj said.

stevep-korg:
OK, I removed the redundant top line of the commit message.

agherzan:
ok to test

PR #306, opened 2018-08-16
machines: use new format for SERIAL_CONSOLES

hhromic:
From the documentation:

    SERIAL_CONSOLES - Provide a value that specifies the baud rate followed by
    the TTY device name separated by a semicolon. Use spaces to separate
    multiple devices

Fixes build error of `systemd-serialgetty` after recent commit 8218bec.
Also, consistently use `ttyAMA0` across all machines.

See: <https://www.yoctoproject.org/docs/current/ref-manual/ref-manual.html#var-SERIAL_CONSOLES>


agherzan:
I agree. Let's separate format from content.

hhromic:
No problem, commit reviewed now for only updating the format.
Regarding the serial device names, indeed my mistake to not test it.
It should be confirmed verified working before committing so will leave it for another time.

PR #312, opened 2018-09-07
RapsbperryPi POE hat and updates

agherzan:
**- What I did**
Updated kernel, firmware and added default support for rpi poe hat.


agherzan:
@floion yes.

christopherhex:
I just tested on a rPI 3B+ with the PoE hat mounted on and can verify that the fan works :-)

PR #313, opened 2018-09-10
When running the RPI3 without vc4graphics linked against libomxil-bel…

wouterlucas:
…lagio.so.0 I get the following error with gst-launch:

```
	Setting pipeline to PAUSED ...
	Pipeline is PREROLLING ...
	Got context from element 'source': gst.soup.session=context, session=(SoupSession)NULL, force=(boolean)false;
	OMX-Cannot open OpenMAX registry file /home/root/.omxregister
	OMX-A Component loader constructor fails. Exiting
	ERROR: pipeline doesn't want to preroll.
	Setting pipeline to NULL ...
```

This is fixed by linking against libopenmaxil.so instead. With that progressive mp4 on gst-lauch works fine.


kraj:
openmaxil.so is precompiled I believe, but lets use it when using userland graphics driver, there is lot of precompiled stuff we use with that setup.

agherzan:
@wouterlucas just a friendly ping for this

wouterlucas:
Sorry for the delays @agherzan I've been traveling and will be traveling for the coming week. I haven't had time to sit down for this yet.

PR #315, opened 2018-09-13
omxplayer: Update and add fixes for OpenSSL 1.1.x and thumb2

kraj:
* Update to latest from upstream
* Backport OpenSSL 1.1.x related patches for ffmpeg
* Pass --cpu based on machine, instead of hardcoding it to arm1176jzf-s
* Make --cc, --cxx and --ld be used from environment, helps building
  with clang
* It requires userland graphics driver to build to encode that in
  COMPATIBLE_HOST


kraj:
ping

agherzan:
@kraj merged. Sorry for the delay.

PR #317, opened 2018-10-02
mesa: Build swrast gallium driver along with vc4

kraj:
when running -ctestimage target, it fails because of errors it finds in
Xorg logs

| Log:
/mnt/a/yoe/build/tmp/work/raspberrypi3-yoe-linux-gnueabi/core-image-sato/1.0-r0/target_logs/Xorg.0.log
| -----------------------
| Central error: [    14.760] (EE) AIGLX error: dlopen of /usr/lib/dri/swrast_dri.so failed (/usr/lib/dri/swrast_dri.so: cannot open shared object file: No such file or directory)

Therefore we enable swrast gallium drivers to make xorg happy


agherzan:
Thanks @kraj . Merged to master.

PR #318, opened 2018-10-03
rpi-config: Use vc4-kms-v3d for raspberrypi3-64 as well

kraj:
vc4-fkms-v3d needs dispmanx, its DRM VC4 V3D driver on top of the dispmanx
display stack, this does not work with 4.14 kernel and since we always
use vc4graphics on 64bit, just keep using vc4-kms-v3d dtbo which should
enable right graphics

Tested with core-image-sato, running glxgears with 55.5fps in
raspberrypi3-64


agherzan:
Merged to master. Thanks, @kraj .

PR #326, opened 2018-11-07
Use needed userland libs with vc4graphics

kraj:
This makes sure the video 3D acceleration is used when using vc4graphics.


kraj:
This will obsolete #313 and #325

PR #335, opened 2018-11-20
CAN: Add Pican2 support

aragua:
In order to make Pican 2 work, we :
- add mcp2515.dto to the overlay list
- add a variable to enable it in local.conf
- create a udev rule to mount interface

http://skpang.co.uk/catalog/pican2-canbus-board-for-raspberry-pi-23-p-1475.html

Signed-off-by: Fabien Lahoudere <fabien.lahoudere@collabora.com>

**- What I did**

**- How I did it**

### DIFF
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -26,6 +26,7 @@ RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
     overlays/iqaudio-dac.dtbo \
     overlays/iqaudio-dacplus.dtbo \
     overlays/lirc-rpi.dtbo \
+    overlays/mcp2515-can0.dtbo \
     overlays/pi3-disable-bt.dtbo \
     overlays/pi3-miniuart-bt.dtbo \
     overlays/pitft22.dtbo \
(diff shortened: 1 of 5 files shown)


aragua:
I can split the patch into several but all changes have the same purpose (enable PIcan) so I propose to use a single commit like "rpi-config: Add ENABLE_CAN for Pican2".
if you prefer several patches, just let me know

agherzan:
`rpi-config: Add ENABLE_CAN for Pican2` means this change only touches the `rpi-config` recipe which is not the case if you use one commit for the entire support. As well, that's exactly what I said. You can either use one commit per component/recipe or use on commit for the entire support but remove the the subject tag,

aragua:
"CAN: " is now dropped from commit title

hhromic:
@agherzan , true, but this is a commit that adds functionality to the layer itself, not to any particular recipe, hence my suggestion. I did this before in these commits:

    meta-raspberrypi: use generic licenses in ${COMMON_LICENSE_DIR} from Poky
    meta-raspberrypi: use `_%` suffix instead of specific versions for bbappends

In my opinion it keeps the commit logs tidy and is easy to spot what is the affected component (the layer itself, some recipe, some conf file, etc).

hhromic:
sure, np! will send a PR with more documentation on the commit log formatting in the coming days

agherzan:
Sounds great. Thanks a lot.

PR #347, opened 2018-12-02
Remove support of the deprecated SDIMG_COMPRESSION variable

koolkhel:
SDIMG_COMPRESSION is not used too much, and it has limitations for
multi-threaded compression process. Instead, meta/classes/image_types.bbclass
does the trick in a generic and supported way, thus it was suggested
to remove SDIMG_COMPRESSION variable and related codepaths altogether
to avoid confusion in the future. IMAGE_FSTYPES variable should be used
instead.

Signed-off-by: Iurii Lunev <koolkhel@mail.ru>

**- What I did**

I removed SDIMG_COMPRESSION definition and usage so no references to that variable exist anymore.

**- How I did it**

By removing :-) See the earlier PR here for reasoning: https://github.com/agherzan/meta-raspberrypi/pull/210

### DIFF
--- a/classes/sdcard_image-rpi.bbclass
+++ b/classes/sdcard_image-rpi.bbclass
@@ -64,13 +64,6 @@ do_image_rpi_sdimg[recrdeps] = "do_build"
 # SD card image name
 SDIMG = "${IMGDEPLOYDIR}/${IMAGE_NAME}.rootfs.rpi-sdimg"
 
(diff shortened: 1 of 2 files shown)


koolkhel:
@agherzan thank you for the tips! I've reworked the commits

agherzan:
Thanks @koolkhel

Issue #324, opened 2018-10-31
omxplayer patch fail on master branch


micha1pu:
**Description**

**Steps to reproduce the issue:**
1. bitbake omxplayer
2.
3.

**Describe the results you received:**

omxplayer-git-r4 do_patch: Command Error: 'quilt --quiltrc /home/studio/build/YOCTO_MASTER/poky-thud/build/tmp/work/arm1176jzfshf-vfp-poky-linux-gnueabi/omxplayer/git-r4/recipe-sysroot-native/etc/quiltrc push' exited with 0  Output:
Applying patch 0001-Remove-Makefile.include-which-includes-hardcoded.patch
patching file Makefile
Hunk #1 FAILED at 1.
1 out of 1 hunk FAILED -- rejects in file Makefile
patching file Makefile.ffmpeg
Hunk #1 FAILED at 1.
1 out of 1 hunk FAILED -- rejects in file Makefile.ffmpeg
The next patch would delete the file Makefile.include,
which does not exist!  Applying it anyway.
patching file Makefile.include
Hunk #1 FAILED at 1.
1 out of 1 hunk FAILED -- rejects in file Makefile.include
Patch 0001-Remove-Makefile.include-which-includes-hardcoded.patch does not apply (enforce with -f)
ERROR: omxplayer-git-r4 do_patch: Function failed: patch_do_patch
ERROR: Logfile of failure stored in: /home/studio/build/YOCTO_MASTER/poky-thud/build/tmp/work/arm1176jzfshf-vfp-poky-linux-gnueabi/omxplayer/git-r4/temp/log.do_patch.17082
ERROR: Task (/home/studio/build/YOCTO_MASTER/poky-thud/meta-raspberrypi/recipes-multimedia/omxplayer/omxplayer_git.bb:do_patch) failed with exit code '1'

**Describe the results you expected:**

**Additional information you deem important (e.g. issue happens only occasionally):**

**Additional details (revisions used, host distro, etc.):**

micha1pu:
In latest omxplayer repository file Makefile.include doesn't exist anymore. Does anybody have same problem?

agherzan:
This sounds strange. @kraj Could this be introduced by the last update? I haven't tried to reproduce.

kraj:
havent seen this here. I will take a look

micha1pu:
i just tried to build on clear system with minimal layers and still same problem, on SUMO its working ok.

bitbake omxplayer
[...]

PR #348, opened 2018-12-03
docs: improve the contributing section

hhromic:
As briefly discussed in PR #335, here are some improvements to the contributing section of the documentation to clarify patch formatting.

* Rephrased some paragraphs
* Better organised the section itself
* Added guideline and formatting examples for commit log messages

Let me know what do you think and suggestions.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -2,37 +2,108 @@
 
 ## Mailing list
 
-The main communication tool we use is a mailing list:
+The main communication tool in use is the Yocto Project mailing list:
+
 * <yocto@yoctoproject.org>
 * <https://lists.yoctoproject.org/listinfo/yocto>
 
-Feel free to ask any kind of questions but always prepend your email subject
-with "[meta-raspberrypi]". This is because we use the 'yocto' mailing list and
-not a perticular 'meta-raspberrypi' mailing list.
+Feel free to ask any kind of questions but please always prepend your email
+subject with `[meta-raspberrypi]` as this is the global *Yocto* mailing
+list and not a dedicated *meta-raspberrypi* mailing list.
 
-## Patches and pull requests
+## Formatting patches
 
-All the contributions should be compliant with the openembedded patch
-guidelines: <http://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines>
+First and foremost, all of the contributions to the layer must be compliant
+with the standard openembedded patch guidelines:
 
-To contribute to this project you should send pull requests to the github mirror
-(<https://github.com/agherzan/meta-raspberrypi>). **Additionally** you can send
-the patches for review to the above specified mailing list.
+* <http://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines>
 
-When creating patches for the mailing list, please use something like:
+In summary, your commit log messages should be formatted as follows:
+
+    <layer-component>: <short log/statement of what needed to be changed>
+
+    (Optional pointers to external resources, such as defect tracking)
+
+    The intent of your change.
+
+    (Optional: if it's not clear from above, how your change resolves
+    the issues in the first part)
+
+    Signed-off-by: Your Name <yourname@youremail.com>
+
+The `<layer-component>` is the layer component name that your changes affect.
+It is important that you choose it correctly. A simple guide for selecting a
+a good component name is the following:
+
+* For changes that affect *layer recipes*, please just use the **base names**
+  of the affected recipes, separated by commas (`,`), as the component name.
+  For example: use `omxplayer` instead of `omxplayer_git.bb`. If you are
+  adding new recipe(s), just use the new recipe(s) base name(s). An example
+  for changes to multiple recipes would be `userland,vc-graphics,wayland`.
+* For changes that affect the *layer documentation*, please just use `docs`
+  as the component name.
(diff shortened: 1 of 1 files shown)


docs/contributing.md
```diff
@@ -2,37 +2,92 @@
...
+
+    <more detailed description of changes and considerations>
+
+    Signed-off-by: Your Name <yourname@youremail.com>
+
+The `<layer-component>` is the layer component name that your changes affect.
+It is important that you choose it correctly. A simple guide for selecting a
+a good component name is the following:
+
+* For changes that affect the *layer itself* such as adding new recipes or
```
kraj: this is not exactly the practice. when adding new recipe you say
```newrecipe: Add version 1.0.0```
so remove that, when changing multiple recipes use
```recipe1,recipe2,recipe3...: Change ...```

File: docs/contributing.md [hunk 2]
```diff
@@ -2,37 +2,92 @@
...
+global functionality, please use `meta-raspberrypi` as the component name.
+* For changes that affect *layer recipes*, please just use the **base name**
+of the recipe as the component name. For example, use `omxplayer` instead of
+`omxplayer_git.bb`. This also applies when adding extra files to recipes.
+* For changes that affect the *layer documentation*, please just use `docs`
+as the component name.
+* For changes that affect *other files*, i.e. under the `conf` directory,
+please use the full path as the component name, e.g. `conf/layer.conf`.
```
kraj: I think this should actually be documented in general openembedded document and rpi layer should point to it. or atleast ensure they are in agreement

docs/contributing.md
```diff
@@ -2,37 +2,92 @@
...
+
+* <https://github.com/agherzan/meta-raspberrypi/issues>
+
+If you push patches that have a GitHub issue associated, please make sure to
+provide the corresponding issue number in the commit log message, just before
+the "Signed-off-by" line(s). For example:
+
+    omxplayer: Forward port patches and fix build
+
+    [Issue #324]
```
kraj: There are github keywords which can change issue states etc. e.g. ```Fixed Issue #324``` etc will reflect correctly.

docs/contributing.md
hhromic: Thanks for the multiple and new recipes suggestion, will reflect.

docs/contributing.md
hhromic: I agree, however the main documentation guidelines are a bit broad. Their own example is:

    Short log / Statement of what needed to be changed.
    
    (Optional pointers to external resources, such as defect tracking)
    
    The intent of your change.
    
    (Optional, if it's not clear from above) how your change resolves the
    issues in the first part.
    
    Tag line(s) at the end.

So I think in the end is up to the individual maintainers to fine-tune how they want their log formatting in their layers. In this case, after the chat with @agherzan we agreed the proposed formatting was handy and made sense to keep things tidy.

We can always try to submit these guidelines upstream, but I think for the moment is good to have it here as the low-hanging fruit for our users. Also, I don't want to turn this into an opinion-wars :)

docs/contributing.md
hhromic: I just used the original tag in the documentation, but yes I agree we should be using the standard GitHub keywords. Will reflect that. Thanks!

docs/contributing.md
agherzan: @kraj I'd love to have a unified guideline but I'm as well worried about the time we need to spend bringing this to oe and discussing a solution at that level. Let's for now fix for us. And if anyone has the time to pursue this to OE, I'll support it as much as I can.

kraj: thats fine, I think, I am interested in keeping it in sync so folks are not confused.

https://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines

and

http://www.openembedded.org/wiki/How_to_submit_a_patch_to_OpenEmbedded

do have detailed prescriptions, now question is should they be improved and simply referred to here with additional github related stuff we talked about or should it rework whats said
in those docs.

hhromic: Yes, the majority of the new content I added comes essentially from these two links as they were in the documentation before I edited it, so I guess is fine :)
The only real addition I would say is the explicit statement of how the commit log messages are expected to be received for this layer (the new guidelines). This essentially is the second option you state (*"should it rework whats said in those docs"*).
I agree that referencing upstream documentation is desirable, but experience also says that users often need to be explicitly and directly told precisely how to do stuff, and in this case is not a big deal to rephrase upstream documentation down here, I think :)

kraj: there is a small isssue with duplicating information. They need to be kept in sync, the OE documents are living documents and they get updated. So we might be signing up for this sync operation instead if we link to those docs then we dont have to worry about it.

hhromic: I agree with that (the commitment of syncing), yes. However I still believe that in this case we are not really changing the upstream docs, we are just specifying more clearly and with examples how things are expected, e.g. the component namings, github keywords, the signed-off tag.

The upstream documentation is broad about that and does not enforce these things so I honestly don't think the improvements in the documentation here are going to get obsolete any time soon.

What you suggest me to do to get this PR going?

hhromic: Hi @kraj, a gently reminder to let me know how you would like me to change the PR to get it done :) Thanks!

agherzan: Merry Christmas, @kraj !

PR #351, opened 2018-12-05
linux-raspberrypi: Backport patch from 4.19 to fix reboot dumps.

drewmoseley:
Some users have reported being hit by this reboot issue when using
Yocto on the Raspberry Pi. This patch is in the 4.19 kernel but
not anything older. Backporting here resolves the reboot dumps.

Signed-off-by: Drew Moseley <drew.moseley@northern.tech>
Tested-by: James E <development@onsolution.com.au>

**- What I did**

**- How I did it**


drewmoseley:
@kraj sure. How do I do that?

kraj:
propose it to
https://github.com/raspberrypi/linux/commits/rpi-4.14.y

drewmoseley:
Doing that now.
FWIW, this is also useful, and cherry-picks cleanly, all the way back to rocko.

drewmoseley:
It looks like this patch is already in the rpi kernel starting with the tag raspberrypi-kernel_1.20180919-1

drewmoseley:
Is there a standard for how far we should bump it? It's not obvious where this layer gets its SRCREV from.

agherzan:
@drewmoseley usually we go to the tip of the branch. We only currently have 4.14 and 4.9 branches supported in this BSP layer.

agherzan:
@jenkins-ag ok to test

drewmoseley:
OK.  https://github.com/agherzan/meta-raspberrypi/pull/352 is a SRCREV update that will include this patch and is likely a preferable fix.

drewmoseley:
No idea if 4.9 needs this fix or not.

PR #370, opened 2019-01-16
Fixes: #169

### LINKED ISSUE #169: Implement  RPI_EXTRA_CONFIG
(putting this in here so I don't forget)

It would be good to have an RPI_EXTRA_CONFIG variable or similar to add arbitrary text to the config file, so you could do something like this:

```
RPI_EXTRA_CONFIG = " \
    # Raspberry Pi 7\" display/touch screen \
    lcd_rotate=2 \
    "
```

memox5:
Enabled escape sequence because on some systems bash doesn't escape sequences by default.

### DIFF
--- a/recipes-bsp/bootfiles/rpi-config_git.bb
+++ b/recipes-bsp/bootfiles/rpi-config_git.bb
@@ -190,7 +190,7 @@ do_deploy() {
     fi
 
     # Append extra config if the user has provided any
-    echo "${RPI_EXTRA_CONFIG}" >> ${DEPLOYDIR}/bcm2835-bootfiles/config.txt
+    printf "${RPI_EXTRA_CONFIG}\n" >> ${DEPLOYDIR}/bcm2835-bootfiles/config.txt
 }
 
 do_deploy_append_raspberrypi3-64() {


memox5:
@agherzan I don't know why a merge commit appeared in PR

memox5:
> 
> 
> Some more things to fix:
> 
>     1. Use "rpi-config" instead of bootfiles.
> 
>     2. fix your Signed-off-by line (name and email)
> 
>     3. `git rebase` on top of master to get rid of the merge commit.

It's the update branch button in the PR that merges, should be fine now, thank you

memox5:
While using -e flag I've come across a strange behavior. In ubuntu 16-08 running on docker, it outputs "-e" in a new line in config.txt, so perhaps there's no common solution for all linux systems after all.

agherzan:
@memox5 what exact base image have you used?

memox5:
@agherzan  I've used ubuntu:16.04

kraj:
Can you see if we can use `printf` instead of `echo`

kraj:
you change is ok but please squash the commits and rebase your branch, so we can get to the final change and forget about the previous iterations.

agherzan:
@memox5 you will probably need a newline if you use printf to maintain the same output as `echo`.

memox5:
@agherzan do you mean like printf "\n${{RPI_EXTRA_CONFIG}" for readability?

agherzan:
`echo` adds a new line at the end of the provided string:

```
$ echo foo
foo
$ printf foo
$ printf "foo\n"
foo
```

memox5:
For example with just a printf command without new lines and the following variable set in local.conf like as stated in the documentation
```
RPI_EXTRA_CONFIG = ' \n\
    # LetsTrust TPM Module \n\
    dtoverlay=letstrust-tpm \n\
    '
```
I get two new lines at the end of config.txt and and empty line between the last entry and my custom configuration lines as expected.

kraj:
you want 
`printf "%s\n" ....`

memox5:
With an "\n" at the end, on docker I ended up with two new lines after the last line and only one on plain ubuntu. I guess because docker build copies files from windows host, some dos2unix trickery is needed there. That might explain the original problem of \n not treated nicely by echo on docker, but on the other end it's known that some linux systems echo doesn't enable escape by default. So does this fix still make sense for you?

kraj:
Its ok to change to using `printf` while echo is simple it can behave different on different shells since its a built-in, however ensure that the change is doing everything that echo was.

memox5:
The problem was echo on some systems wasn't escaping \n sequence in the variable, printf does. Can we safely assume that in the RPI_EXTRA_CONFIG variable there shouldn't be any unknowingly escaped sequences?

agherzan:
@jenkins-ag ok to test

PR #380, opened 2019-02-11
qtbase: fix xkbcommon-evdev packageconfig for 5.12

sashko:
Patch that made `meta-qt5` use this feature has been reverted for `5.12`,
which is used by `thud` layer of `meta-qt5` with commit `33c578729e721ac80e`.

Signed-off-by: Oleksandr Kravchuk <oleksandr.kravchuk@pelagicore.com>


sashko:
> I am not sure where this was reverted. Can you add few links ?

https://github.com/meta-qt5/meta-qt5/blob/master/recipes-qt/qt5/qtwayland/0001-Revert-use-new-feature-name-xkbcommon_evdev-xkbcommo.patch

kraj:
I think the qt patch should be backported to thud for better can you please submit a backport request to meta-qt5

agherzan:
I'm not sure I understand what is going on here. Can someone explain?

kraj:
> I'm not sure I understand what is going on here. Can someone explain?

this fix wont be needed on thud if a patch in thud can be backported from master in meta-qt5 itself.
so I am asking for backporting the needed patch in meta-qt5 itself instead of adjust meta-rpi bbappend.

PR #383, opened 2019-02-13
linux-raspberrypi: Update to 4.14.98

majorz:
**- What I did**
Updated RaspberryPi kernel and firmware.

### DIFF
--- a/recipes-bsp/common/firmware.inc
+++ b/recipes-bsp/common/firmware.inc
@@ -1,11 +1,11 @@
-RPIFW_DATE ?= "20181211"
-SRCREV ?= "de8589ef09567f12508faedc8ced3046b078e9b5"
+RPIFW_DATE ?= "20190212"
+SRCREV ?= "83977fe3b6ef54c1d29c83b0a778d330f523441f"
 RPIFW_SRC_URI ?= "https://github.com/raspberrypi/firmware/archive/${SRCREV}.tar.gz"
 RPIFW_S ?= "${WORKDIR}/firmware-${SRCREV}"
 
 SRC_URI = "${RPIFW_SRC_URI}"
-SRC_URI[md5sum] = "077b7bff0c35f9a0b9b0978dd4f221cb"
-SRC_URI[sha256sum] = "748f6443134abad39ec9d57423b19d219f51fc362500bb4760fe415520b4fbe5"
+SRC_URI[md5sum] = "5ccdb5447cbdd3ee0158a514f7b76cb9"
+SRC_URI[sha256sum] = "9a34ccc4a51695a33206cc6c8534f615ba5a30fcbce5fa3add400ecc6b80ad8a"
 
 
 PV = "${RPIFW_DATE}"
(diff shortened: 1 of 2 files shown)


kraj:
it will be good to get a  URL to changelog between these two commits. So one can look into the changes that it brings in.

majorz:
@kraj @agherzan I was thinking about the Changelog type of information and here is what I came up with.

For a particular release, e.g. 4.14.98, the list of the RPi specific changes can be found here: https://github.com/raspberrypi/firmware/commit/83977fe3b6ef54c1d29c83b0a778d330f523441f

Linux stable 4.14.98 is announced here: https://lwn.net/Articles/779133/

Including those two links seems sufficient.

Here in particular the upgrade is from 4.14.87 to 4.14.98, which is a bigger gap. That lead me to think that maybe meta-raspberrypi needs to be synced with each specific patch version. And more specifically synced with what is released in the raspberrypi/firmware repo - so synced with both the kernel and firmware. Some simple automation could be helpful.

What do you guys think?

agherzan:
@kraj It is not an easy one. linux-rpi doesn't maintain (afaik) a clean changelog that we can reference. Most of this kind of information resides in git log.

agherzan:
@majorz I'm not sure how relevant that is because it doesn't provide the changelog *from* the version we update. It is only the changelog since the last release.

agherzan:
@jenkins-ag add to whitelist

agherzan:
@jenkins-ag ok to test

majorz:
@agherzan right, I did not phrased my message well. My point was that it will be easier to show changes if the updates on meta-raspberrypi are done incrementally. Currently the gap is 10 patch versions - that would be 20 links in total. 2x10 for the RPi kernel links and 2x10 for the LWN announcements.

My main question is not actually about the changelog, but whether the RPi kernel and the RPi firmware should be synced simultaneously. Right now I only PRed the kernel bump, but not the firmware. Could there be a possible breakage if those are not synced together?

agherzan:
I usually update them in tandem. We've seen in the past breakage in that regards but not lately. I'd say it's not a huge risk now but I'd rather have the firmware updated too. It's a good idea anyway. @majorz

majorz:
@agherzan thanks, should I amend the current commit or add a new one with the firmware update?

agherzan:
@majorz you can make it part of the same PR but different commit.

majorz:
Cool, on it.

majorz:
@agherzan It is done now.

kraj:
> @kraj It is not an easy one. linux-rpi doesn't maintain (afaik) a clean changelog that we can reference. Most of this kind of information resides in git log.

one can use something like this compare view github URL in commit

`https://github.com/$USER/$REPO/compare/$REV_A...$REV_B`

which for this change would mean

https://github.com/raspberrypi/linux/compare/faf0452f03b09c7063a2b14b23f61b3bd7eb264a...5d63a4595d32a8505590d5fea5c4ec1ca79fd49d

agherzan:
@kraj Even that gets truncated for that amount of changes but yeah, maybe that would work.

kraj:
yes thats ok. All I am asking is to add the link to commit msg. so one can use that link
to go to github changes.

PR #386, opened 2019-02-18
CM3 improvements and tests

agherzan:
**- What I did**
Make the cm3 machine stand alone and test core image basic with cm3 and cm3+ in both with uboot and without configurations


kraj:
What does it get us on top of normal rpi3 config

agherzan:
@kraj different default serial console and no wifi/bt firmware.

PR #388, opened 2019-02-19
Thud backports from master

agherzan:


agherzan:
Pushed the CM3 changes and removed the `9cc89cb raspberrypi3-64.conf: Use more appropriate tune` as it's not available in `thud`.

PR #406, opened 2019-04-02
layer.conf: Add warrior to compatible release series

kraj:


shr-project:
Is someone still using master branch of meta-raspberrypi with sumo/thud? Shouldn't the older releases be removed once the warrior release settles a bit?

kraj:
Yes, although it should be done sequentially, meaning send a patch to remove older branches as a followup once warrior is working ok

kraj:
@agherzan this needs to be merged asap

agherzan:
Why do we want to remove old branches? I'm pretty sure there are still people working on older yocto releases.

kraj:
Idea is that they are then forced to use the right branch but lets do that as a follow up.

PR #412, opened 2019-04-17
Fix different signatures for different rpi MACHINEs which should share the same package feed

shr-project:
Fixes for issues discussed here:
https://github.com/agherzan/meta-raspberrypi/commit/1317a894faa671c82193c1f18074d84db2cb9e35#commitcomment-33197660


agherzan:
Looks good. Merged to master.

PR #417, opened 2019-05-17
meta-raspberrypi: Enabling vc4graphics machine feature.

kalyan-nagabhirava:
gstreamer1.0-plugins-bad : Packaging gstreamer
plugins.

Signed-off-by: Riyaz <Riyaz.l@ltts.com>

### DIFF
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -70,7 +70,7 @@ KERNEL_IMAGETYPE_DIRECT ??= "zImage"
 KERNEL_IMAGETYPE ?= "${@bb.utils.contains('RPI_USE_U_BOOT', '1', \
         '${KERNEL_IMAGETYPE_UBOOT}', '${KERNEL_IMAGETYPE_DIRECT}', d)}"
 
-MACHINE_FEATURES += "apm usbhost keyboard vfat ext2 screen touchscreen alsa bluetooth wifi sdio"
+MACHINE_FEATURES += "apm usbhost keyboard vfat ext2 screen touchscreen alsa bluetooth wifi sdio vc4graphics"
 
 # Raspberry Pi has no hardware clock
 MACHINE_FEATURES_BACKFILL_CONSIDERED = "rtc"


Khan3033:
> Please divide this into two separate patches. And I would like to gather more information on before switching default to vc4graphics on 32bit.

We have addressed your comments kindly review, thanks.

kraj:
switching to 4.19 is being addressed in https://github.com/agherzan/meta-raspberrypi/pull/418
so make this pull just to swich the graphics driver, and we should make it across all rpis and just not for rpi3 alone.

conf/machine/raspberrypi3.conf
```diff
@@ -13,6 +13,8 @@ MACHINE_EXTRA_RRECOMMENDS += "\
     bluez-firmware-rpidistro-bcm4345c0-hcd \
 "
 
+MACHINE_FEATURES += "vc4graphics"
```
kraj: this change should be make common across all rpi machine, rest of the change should be dropped from this patch.

Khan3033: Hi @kraj ,
Committed as per the change, Kindly review,
9cd3dc794461ccfb6c1354d16a8fc7deabe53c3d

psivasubramanian:
Hi @kraj @agherzan 
We hope your review comments are addressed. Please verify and let us know if any comments.

agherzan:
@psivasubramanian I'm not sure how mu comment is addressed. Maybe it was missed:

> The commit on this PR mentions nothing about why this is needed. What is the issue it addresses? Why do we want to enable by default vc4 for RaspberryPi 3? Is this part of a conversation I'm missing the context from? But even so, the summary of it should be available in the commit log.

This changes now a little by making the question non-rpi3 specific but the core of it still stands.

agherzan:
I am tempted to consider this PR stale. I'll close it soon if I get no feedback.

psivasubramanian:
> @psivasubramanian I'm not sure how mu comment is addressed. Maybe it was missed:
> 
> > The commit on this PR mentions nothing about why this is needed. What is the issue it addresses? Why do we want to enable by default vc4 for RaspberryPi 3? Is this part of a conversation I'm missing the context from? But even so, the summary of it should be available in the commit log.
> 
> This changes now a little by making the question non-rpi3 specific but the core of it still stands.

Sorry for the delayed reply. We thought the commit log (as below) answered your question to some extent. But it seems it doesn't.

meta-raspberrypi: Enabling open-source vc4graphics driver for all RPI platforms
Appending vc4graphics machine feature to use vc4graphics driver.

We are in the process of evaluating the performance of vc4graphics against userland for providing justification to this PR. Does that will be suffice?

agherzan:
@psivasubramanian I've never received that message and the commit log just translates the commit content. There is not explanation whatsoever. Even so, you can always customize your build to switch on vc4. Isn't that working for you?

kraj:
> @psivasubramanian I've never received that message and the commit log just translates the commit content. There is not explanation whatsoever. Even so, you can always customize your build to switch on vc4. Isn't that working for you?

@agherzan with linux kernel 4.19+ vv4graphics is in pretty good shape, so switching it to be default may be a compelling option for 32bit builds as well, we already use it by default for 64bit.

psivasubramanian:
Hi @agherzan 
Hope this commit log (https://github.com/agherzan/meta-raspberrypi/pull/417/commits/680b89471dde76b5e9650e3865be8d01911faf69) gives more information about this change as well as answers your question. Kindly let us know if any comments/thoughts.

agherzan:
@kraj We do but we will need some testing on 32B as well. It doesn't feel right to switch just like that. What about building the support but don't activating the overlay by default? My reason behind being a little reticent about this is because while working on RPi4 I had a hard time working with vc4 and finally managed to only make it work with fkms. CC @psivasubramanian

kraj:
> @kraj We do but we will need some testing on 32B as well. It doesn't feel right to switch just like that. What about building the support but don't activating the overlay by default? My reason behind being a little reticent about this is because while working on RPi4 I had a hard time working with vc4 and finally managed to only make it work with fkms. CC @psivasubramanian

fkms issue was a regression in kernel which should be fixed with
https://github.com/agherzan/meta-raspberrypi/pull/436

So we should drop fkml for rpi4 as well after above is merged.

vc4 is where things are headed to, mesa works equally well on both 32bit and 64bit and
with 4.19 it also supports v4l2, so I think we are better off supporting it as default now a days
we can always revert or define another machine type something like
`rpi3-vc4graphics` and so on

agherzan:
@kraj you are the main point of control in that area so I'll take your advice. Let's see who screams.

agherzan:
@kraj can you also update your review, please?

psivasubramanian:
@kraj @agherzan Rebased onto latest. Please review.

agherzan:
@kraj @psivasubramanian I had to rebase master because the log of this commit was really ugly which I haven't caught at review this time.

PR #419, opened 2019-05-22
rpi-default-providers.inc: default provider for virtual/kernel_poky-tiny

quitschbo:
For the distro poky-tiny a provider for virtual/kernel_poky-tiny
is needed instead of virtual/kernel.
This was missing in rpi-default-providers.inc

Signed-off-by: Michael Weiß <michael.weiss@aisec.fraunhofer.de>

**- What I did**

**- How I did it**

### DIFF
--- a/conf/machine/include/rpi-default-providers.inc
+++ b/conf/machine/include/rpi-default-providers.inc
@@ -1,6 +1,7 @@
 # RaspberryPi BSP default providers
 
 PREFERRED_PROVIDER_virtual/kernel ?= "linux-raspberrypi"
+PREFERRED_PROVIDER_virtual/kernel_poky-tiny ?= "linux-raspberrypi"
 PREFERRED_PROVIDER_virtual/xserver ?= "xserver-xorg"
 PREFERRED_PROVIDER_virtual/egl ?= "${@bb.utils.contains("MACHINE_FEATURES", "vc4graphics", "mesa", "userland", d)}"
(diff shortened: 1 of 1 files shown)


conf/machine/include/rpi-default-providers.inc
```diff
@@ -1,6 +1,7 @@
 # RaspberryPi BSP default providers
 
 PREFERRED_PROVIDER_virtual/kernel ?= "linux-raspberrypi"
+PREFERRED_PROVIDER_virtual/kernel_poky-tiny ?= "linux-raspberrypi"
```
kraj: This is a distro specific override, I wonder why is this needed, The right fix it to change it in meta-poky from = to ?=
see
https://git.yoctoproject.org/cgit/cgit.cgi/meta-yocto/tree/meta-poky/conf/distro/poky-tiny.conf#n40

quitschbo: well i know that. Also you can set the provider in local.conf. However other layers such as meta-intel do the same in their machine config, providing out-of-the-box experience when compiling poky-tiny.

kraj: If this was fixed in distribution layer as suggested then bsp and other layers don’t have to do it meta-intel doing it doesn’t give legitimacy to do it elsewhere

quitschbo: Yeah I'm convinced. So I will try to get it upstream in poky.

PR #420, opened 2019-05-22
mesa-gl: Fix build after poky switched to meson

agherzan:
Signed-off-by: Andrei Gherzan <andrei@gherzan.ro>

### DIFF
--- a/recipes-graphics/mesa/mesa-gl/0001-Allow-enable-DRI-without-DRI-drivers.patch
+++ b/recipes-graphics/mesa/mesa-gl/0001-Allow-enable-DRI-without-DRI-drivers.patch
@@ -1,0 +1,45 @@
+From 8e0a58c8eea7ed70071cac139655700fdfa16445 Mon Sep 17 00:00:00 2001
+From: Andrei Gherzan <andrei@gherzan.ro>
+Date: Wed, 22 May 2019 18:32:07 +0100
+Subject: [PATCH] Allow enable DRI without DRI drivers
+
+Upstream-status: Pending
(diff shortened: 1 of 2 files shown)


agherzan:
@kraj I've seen you've sent it already as part of http://lists.openembedded.org/pipermail/openembedded-core/2019-May/282723.html . How soon should we expect it merged?

kraj:
It’s already in master next

agherzan:
The PRs are tested against master so we will have to wait for that.

kraj:
No problem but I think CI should pick all other layers to a known version accept meta-raspberry pi in my opinion otherswise you are also running CI for them :)

agherzan:
@kraj That is somehow true. The problem is that here and there we do depend on the core so we want to make sure that as dev advances we catch up with incompatible changes we need to make. All the PRs against specific branches are tested against the same branch on the other layers. But how to choose the dependency revisions for the master PR in such a way that we can catch things we need to change but without fixing what they break.

kraj:
I therefore like the concept of monorepo, Like we have done with YoeDistro

https://github.com/YoeDistro/yoe-distro

we lock all layers and move them in a lock step.

agherzan:
That is something I was thinking about as well. It would help this for the cost of having to maintain the submodules at a per branch level. Which is not insane but it is more work while this approach works as it is and I only saw twice this situation happening. I'll keep an eye on it. If this becomes an annoyance I'll migrate to submodules.

agherzan:
Fixed in master oe-core so closing this.

PR #423, opened 2019-06-04
meta-raspberrypi: sdcard_image-rpi, Minor bug in use of FATPAYLOAD variable

malus-brandywine:
FATPAYLOAD has been put into double quotation marks. Otherwise, when its value is compound  (like "filename_1   filename_2   filename_3") it generates parsing error.

Signed-off-by: Nataliya Korovkina <malus.brandywine@gmail.com>

### DIFF
--- a/classes/sdcard_image-rpi.bbclass
+++ b/classes/sdcard_image-rpi.bbclass
@@ -143,7 +143,7 @@ IMAGE_CMD_rpi-sdimg () {
         fi
     fi
 
-    if [ -n ${FATPAYLOAD} ] ; then
+    if [ -n "${FATPAYLOAD}" ] ; then
         echo "Copying payload into VFAT"
         for entry in ${FATPAYLOAD} ; do
             # add the || true to stop aborting on vfat issues like not supporting .~lock files


malus-brandywine:
> Looks good. On tiny thing you need to address is commit log. Please check https://github.com/agherzan/meta-raspberrypi/blob/master/docs/contributing.md for more details.

Updated.

malus-brandywine:
Hi Andrei, I sent the second patch for the same file. There's additional small option implemented.

agherzan:
Thanks @malus-brandywine. Can you tell me exactly what were the files you had issues with? I'm asking because different components rely on a filename. So if we miss symlink we might even brick the boot.

malus-brandywine:
Xen hypervisor installs files in /boot folder of rootfs image, which is [the folder] mount point for fat32 partition, so everything installed there got hidden. For now I should explicitly list those files in FATPAYLOAD so they get copied to fat32 partition.
Regarding this patch, by default, all files produced by other packages and placed in rootfs' /boot will be copied, if unconditionally listed in FATPAYLOAD. I wanted to have an option to avoid duplicating. Of course, recipes which are set to skip symlinks should consider which files left available.
You might ask why not to control what goes into FATPAYLOAD in the first place instead of filtering out files later. I thought that it's more convenient to have one recipe which keeps complete list of files unchanged while dependant recipes can use the option to filter symlinks out.

agherzan:
I see now @malus-brandywine . What layers are you using? I'm asking because this doesn't seem to be BSP specific issue. We don't set in this layer any mountpoint for the boot partition.

malus-brandywine:
@agherzan , my previous description was wrong. Xen binary (and symlinks) are installed in /boot folder of rootfs partition and don't go to FAT partition. However, they still can be reachable for u-boot. So, it's a matter of convenience, not necessity to have them in FAT partition. So, I withdraw my second patch.

However, my first patch is still valid if the code keeps supporting FATPAYLOAD as a list of files.

agherzan:
@malus-brandywine In that case, can you please rebase the branch removing second commit?

malus-brandywine:
Done.

agherzan:
@jenkins-ag ok to test

PR #438, opened 2019-07-15
linux-raspberrypi: Build dtbs with dtbs make target for rpi4-64

agherzan:
We already do this for rpi3-64 and we will need it for rpi4-64 as well.
See 6c4de0b5fe44b8e661f1391ee8540a7f04d75315 for more details.

Signed-off-by: Andrei Gherzan <andrei@gherzan.ro>

CC @diegosueiro


agherzan:
I like that. It will help reusing code too.

PR #442, opened 2019-07-18
Check some config values (ex: ENABLE_I2C) against "1" instead of stri…

Carton32:
**- What I did**

I changed the check for some variables in the rpi-config_git.bb.

When we read the docs, we have the feelings that theses variables are boolean ones. So I was setting, for example in my distro.conf file the variable ' ENABLE_I2C = "1" ' to enable I2C. Then I wanted to disable it by simply setting 'ENABLE_I2C' to "0" but it wasn't working. So I noticed that, for example, ' ENABLE_UART ' was checked with ' = "1" ' condition and some other "boolean" was checked against ' -n ' like for ENABLE_I2C.

This commit tries to have an uniform behavior for all variables that are shown in the doc under the format ' VARIABLE = "1" ' to enable them and the reader can think they are kind of 'boolean' values.


agherzan:
@jenkins-ag ok to test

agherzan:
Thanks @Carton32 . Merged to master.

agherzan:
Facepalm. I missed the commit log. It doesn't include signed of by and the subject format is all in one line.

PR #447, opened 2019-07-19
Armstubs fixes

agherzan:


agherzan:
@kraj there is only one per configuration. That's why the singular there.

kraj:
> @kraj there is only one per configuration. That's why the singular there.

sounds good.

PR #448, opened 2019-07-19
Warrior backports

agherzan:


agherzan:
@kraj most of the functional changes are a dependency to rpi4. The rest are just minor fixes. Do you think we should filter out some stuff?

kraj:
I don’t think so on the contrary we should keep master working with many releases

Issue #456, opened 2019-07-28
rpi-test-image unbuildable because of crda (new issue?)


MarcoBorrini99:
**Description**

With lastest master, bitbake fails to build rootfs for a Rpi4 rpi-test-image because of crda.

**Steps to reproduce the issue:**
1. clone meta-raspberrypi master branch
2. clone meta-openembedded master branch
3. create building env:
```
[marco@Port371]: ~/src/poky>$ source oe-init-build-env rpi-build
```
4. edit conf/local.conf adding:
```
MACHINE ??= "raspberrypi4"
PACKAGE_CLASSES ?= "package_ipk"
```
5. edit conf/bblayers.conf as:
```
BBLAYERS ?= " 							\
  /home/marco/src/poky/meta 					\
  /home/marco/src/poky/meta-poky 				\
  /home/marco/src/poky/meta-yocto-bsp 				\
  /home/marco/src/poky/meta-openembedded/meta-oe 		\
  /home/marco/src/poky/meta-openembedded/meta-multimedia 	\
  /home/marco/src/poky/meta-openembedded/meta-networking 	\
  /home/marco/src/poky/meta-openembedded/meta-python 		\
  /home/marco/src/poky/meta-raspberrypi				\
  "
```
6. start buildig image
`[marco@Port371]: ~/src/poky/rpi-build>$ bitbake rpi-test-image`

**Describe the results you received:**
```
[marco@Port371]: ~/src/poky/rpi-build>$ bitbake rpi-test-image
WARNING: Host distribution "arch" has not been validated with this version of the build system; you may possibly experience unexpected failures. It is recommended that you use a tested distribution.
Loading cache: 100% |#################################################################################################| Time: 0:00:00
Loaded 3381 entries from dependency cache.
NOTE: Resolving any missing task queue dependencies

Build Configuration:
BB_VERSION           = "1.43.0"
BUILD_SYS            = "x86_64-linux"
NATIVELSBSTRING      = "universal"
TARGET_SYS           = "arm-poky-linux-gnueabi"
MACHINE              = "raspberrypi4"
DISTRO               = "poky"
DISTRO_VERSION       = "2.7+snapshot-20190728"
TUNE_FEATURES        = "arm vfp cortexa7 neon vfpv4 thumb callconvention-hard"
TARGET_FPU           = "hard"
meta                 
meta-poky            
[...]

PR #458, opened 2019-07-30
packagegroup-rpi-test: Depend on wireless-regdb instead of crda

kraj:
this should bring in crda if needed

Fixes issue #456


agherzan:
Thanks @kraj

PR #462, opened 2019-08-05
xorg-xserver: Adapt bbappend to latest OE-core

kraj:
There is no need to override complete packageconfig but just the openGL
part, this ensures that any common change in oe-core reflects for rpi as
well and does not cause the packageconfigs to go stale


agherzan:
Thanks @kraj

PR #470, opened 2019-08-12
python-rtimu,python-sense-hat: Convert to py3 modules

kraj:
py2 is on its way out from oe-core


agherzan:
Thanks @kraj . Merged to master.

PR #475, opened 2019-08-15
mesa: Enable vc4, v3d, kmsro gallium drivers

kraj:
This enables new v3d driver which needs kmsro as dependency

This will start working fully after oe-core change [1] is merged

[1] http://lists.openembedded.org/pipermail/openembedded-core/2019-August/285763.html


kraj:
@agherzan this can be merged now that OE-Core change is merged too

PR #493, opened 2019-09-11
raspberrypi-linux: removed protocoll setting from SRC_URI

### LINKED ISSUE #465: Use https protocol instead of git
The ssh port for git may be blocked on some setups. Using https would be more compatible.

https://github.com/agherzan/meta-raspberrypi/blob/9d0935c9bb1309431e62b8f8341eb503653e5ff5/recipes-kernel/linux/linux-raspberrypi_4.19.bb#L8

saeugetier:
Default protocoll for git will be used. Default setting is in the most case https.

Signed-off-by: Timm Eversmeyer saeugetier@gmail.com

fixes #465


agherzan:
@jenkins-ag ok to test

PR #501, opened 2019-10-08
bcm2835-bootfiles: tighten up dep on rpi-config

mostthingsweb:
Intuitively, bcm2835-bootfiles:do_deploy should depend on rpi-config:do_deploy.

This indirectly resolves a missing dep between rpi-config:do_deploy and
do_image_rpi_sdimg (on an image recipe inheriting from sdcard_image-rpi.bbclass).
This manifested as changes to rpi-config (e.g. setting or unsetting
ENABLE_* variables) not triggering a rebuild of the SD card image.

Signed-off-by: Chris Laplante <mostthingsweb@gmail.com>

### DIFF
--- a/recipes-bsp/bootfiles/bcm2835-bootfiles.bb
+++ b/recipes-bsp/bootfiles/bcm2835-bootfiles.bb
@@ -34,6 +34,8 @@ do_deploy() {
     touch ${DEPLOYDIR}/${PN}/${PN}-${PV}.stamp
 }
 
+do_deploy[depends] += "rpi-config:do_deploy"
+
 addtask deploy before do_build after do_install
 do_deploy[dirs] += "${DEPLOYDIR}/${PN}"
 
--- a/classes/sdcard_image-rpi.bbclass
+++ b/classes/sdcard_image-rpi.bbclass
@@ -55,6 +55,7 @@ do_image_rpi_sdimg[depends] = " \
     dosfstools-native:do_populate_sysroot \
     virtual/kernel:do_deploy \
     ${IMAGE_BOOTLOADER}:do_deploy \
+    rpi-config:do_deploy \
     ${@bb.utils.contains('MACHINE_FEATURES', 'armstub', 'armstubs:do_deploy', '' ,d)} \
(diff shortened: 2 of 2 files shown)


recipes-bsp/bootfiles/bcm2835-bootfiles.bb
```diff
@@ -34,6 +34,8 @@ do_deploy() {
     touch ${DEPLOYDIR}/${PN}/${PN}-${PV}.stamp
 }
 
+do_deploy[depends] += "rpi-config:do_deploy"
```
agherzan: I agree that the dependency is not clear enough but I'm actually thinking that this is not the actual consumer of the `config.txt`. The actual consumer is the image generation task `do_image_rpi_sdimg`. It would sound a little more reasonable to have this dependency on that [side](https://github.com/agherzan/meta-raspberrypi/blob/d34e054d5140792150d2a7ae3c0264eb254c8aa3/classes/sdcard_image-rpi.bbclass#L52). What do you think?

mostthingsweb:
I think it would make sense to have it in both places. If someone does bcm2835-bootfiles:do_deploy, then (IMO) rpi-config:do_deploy should still run. I see rpi-config as an implementation detail of bcm2835-bootfiles.

recipes-bsp/bootfiles/bcm2835-bootfiles.bb
mostthingsweb: I think it would make sense to have it in both places. If someone does bcm2835-bootfiles:do_deploy, then (IMO) rpi-config:do_deploy should still run. I see rpi-config as an implementation detail of bcm2835-bootfiles.

I have just committed a change that does this.

(Sorry for duplicate comment, I forgot to use Reply)

agherzan:
@agherzan-ci ok to test

agherzan:
@mostthingsweb I'm alright with this.

agherzan:
Merged to master. Thanks.

mostthingsweb:
My bad, shouldn't have clicked Close

PR #511, opened 2019-10-21
Licensing clarifications

pbrkr:
The 'Proprietary' license is used for recipes where we can't redistribute the sources. The archiver bbclass excludes recipes with this license when collecting recipe sources so we should only use it where we absolutely can't redistribute anything.

The sources downloaded by the bcm2835-bootfiles and vc-graphics recipes are pre-compiled binaries which we are allowed to redistribute as long as it's for use on a Raspberry Pi device. I've given this license the name 'Broadcom-RPi' and placed it in the custom-licenses directory.

The armstubs source files were found to have a 3-clause BSD license header.

The 'Broadcom' license in the custom-licenses directory was found not to be used. The name is a bit too generic anyway, it's easy to accidentally clash if some other layer adds a recipe with a different license from Broadcom and chooses to just call it 'Broadcom'.


agherzan:
@jenkins-ag ok to test

agherzan:
@agherzan-ci ok to test

agherzan:
@agherzan-ci add to whitelist

agherzan:
Thanks @pbrkr . Merged to master.

PR #525, opened 2019-11-13
Fix u-boot support for rpi4

lbonn:
u-boot stopped working on rpi4 starting from the following commit
included in 2019.10: https://gitlab.denx.de/u-boot/u-boot/commit/9de5b89e4c898ec770878eb4848588c635a37bac

According to the mailing list, it has been since fixed on u-boot master: https://www.mail-archive.com/u-boot@lists.denx.de/msg346771.html

After testing, it seems to be accurate!

### DIFF
--- a/recipes-bsp/u-boot/u-boot-2019.10.inc
+++ b/recipes-bsp/u-boot/u-boot-2019.10.inc
@@ -1,0 +1,8 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/u-boot-2019.10:"
+
+SRC_URI_append_rpi = " \
+    file://0001-libfdt-fdt_address_cells-and-fdt_size_cells.patch \
+    file://0002-libfdt-return-correct-value-if-size-cells-property-i.patch \
+    file://0003-libfdt-Allow-size-cells-of-0.patch \
+    file://0004-dm-Fix-default-address-cells-return-value.patch \
+"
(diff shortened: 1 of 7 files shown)


pbrkr:
This looks good to me. Quick question though - is it just Raspberry Pi devices which are affected? If not maybe this could be upstreamed to oe-core.

lbonn:
That's a good point and I cannot say for sure but I think the regression is really specific to rpi4.

Clarification, after a closer look: the problem was known at the time the offending patch was proposed https://lists.denx.de/pipermail/u-boot/2019-September/382938.html. So, things were merged in the wrong order and 2019.10 was released in the middle.

Not really opposed to send it to oe-core but not sure how to sell it there.

pbrkr:
Ok cool - I think that's clear enough for us to carry it here.

recipes-bsp/u-boot/u-boot-2019.10.inc
```diff
@@ -0,0 +1,8 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/u-boot-2019.10:"
+
+SRC_URI_append = " \
```
pbrkr: We should probably make this `SRC_URI_append_rpi` just in case it affects other machines.

lbonn: yes, updated

lbonn:
updated the commit message

agherzan:
@agherzan-ci ok to test

lbonn:
Actually, it looks like poky is gonna update u-boot to 2020.01-rc3 soon (in master-next right now: https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/?h=master-next&id=e8e21951c707938bb5228838df38681f9e51d11d). Will spin a test in the meantime.

So, you might want to hold on on this.

lbonn:
u-boot in master now contains the fix: https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/?id=16a39715d9d07cf847d433f5a3c14c7374ea5a6e.

No need to merge this anymore.

agherzan:
Thanks, @lbonn for chasing this up.

PR #537, opened 2019-12-10
Updates

agherzan:
Update kernel and firmware


agherzan:
@kraj Makes sense. Pushed.

agherzan:
@kraj just a small ping here so we can merge this.

PR #541, opened 2019-12-20
linux-firmware-rpidistro_git: Update wifi firmware

floion:
This updates brcmfmac43455-sdio.bin to version 7.45.189
and brcmfmac43430-sdio.bin to version 7.45.98.83

This updated upstream firmware also includes the changes from
0001-brcmfmac43455-sdio.txt-Follow-raspbian-change-for-bo.patch
so we remove this patch along with removing the do_clean_pc
function which is not needed anymore since we don't apply
any other patches to the upstream source.

Signed-off-by: Florin Sarbu <florin@balena.io>

**- What I did**

**- How I did it**


floion:
Thank you

agherzan:
@floion is this a backport?

PR #549, opened 2020-01-06
rpi cm3: migrate KBUILD DEFCONFIG to linux-raspberrypi

ricardosalveti:
KBUILD_DEFCONFIG should be defined by the kernel recipe, not board config (otherwise it will be set on every other kernel recipe that gets used by the build system).


agherzan:
@agherzan-ci ok to test

agherzan:
Merged to master. Thanks, @ricardosalveti .

PR #560, opened 2020-01-15
[RFC] kas-poky-rpi: add kas file

texierp:
The kas tool provides an easy mechanism to setup bitbake based projects.

As explained in [1] the key features provided by kas are:

- clone and checkout bitbake layers
- create default bitbake settings (machine, arch, ...)
- launch minimal build environment, reducing risk of host contamination
- initiate bitbake build process

So, let's create a minimal example in this layer to build an image with a
small effort. This seems interesting for new users of the meta-raspberrypi.

For example:

$: sudo pip3 install kas
$: mkdir rpi
$: cd rpi
$: git clone https://github.com/agherzan/meta-raspberrypi.git
$: kas build meta-raspberrypi/kas-rpi.yml

and that's all !

To adjust the construction with specific options (I2C, SPI, ...), simply add
a section as follows:

local_conf_header:
  rpi-specific: |
    ENABLE_I2C = "1"
    RPI_EXTRA_CONFIG = "dtoverlay=pi3-disable-bt"

To configure the machine, the user have to update the 'machine' variable.
And the same for the 'distro'.

[1] - https://kas.readthedocs.io/en/1.0/userguide.html

Signed-off-by: Pierre-Jean Texier <pjtexier@koncepto.io>

### DIFF
--- a/kas-poky-rpi.yml
+++ b/kas-poky-rpi.yml
@@ -1,0 +1,58 @@
+header:
+  version: 8
+
+machine: raspberrypi4
+distro: poky
+target:
+  - core-image-base
+
+repos:
+  meta-raspberry:
+
+  poky:
+    url: https://git.yoctoproject.org/git/poky
+    refspec: master
+    layers:
+      meta:
+      meta-poky:
+      meta-yocto-bsp:
+
+  meta-openembedded:
+    url: http://git.openembedded.org/meta-openembedded
+    refspec: master
+    layers:
+      meta-oe:
+      meta-python:
+      meta-networking:
+      meta-perl:
+
+  meta-qt5:
+    url: https://github.com/meta-qt5/meta-qt5/
+    refspec: master
+
+bblayers_conf_header:
+  standard: |
+    POKY_BBLAYERS_CONF_VERSION = "2"
+    BBPATH = "${TOPDIR}"
+    BBFILES ?= ""
+local_conf_header:
+  reduce_diskspace: |
+    INHERIT += "rm_work_and_downloads"
+  standard: |
+    CONF_VERSION = "1"
+    PACKAGE_CLASSES = "package_rpm"
+    SDKMACHINE = "x86_64"
+    USER_CLASSES = "buildstats image-mklibs image-prelink"
+    PATCHRESOLVE = "noop"
+  debug-tweaks: |
+    EXTRA_IMAGE_FEATURES = "debug-tweaks"
+  diskmon: |
+    BB_DISKMON_DIRS = "\
+        STOPTASKS,${TMPDIR},1G,100K \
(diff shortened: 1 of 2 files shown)


pbrkr:
I like this idea, it will make testing a little easier. I'd recommend the following changes though:

* Put 'poky' in the name of the kas file so we know this uses the poky distro, maybe "kas-poky-rpi.yml".
* Add the example of how to use this to the readme, maybe in a section "Quick start with kas" after the existing "Quick start" section.
* Re-word the commit summary slightly to begin with "kas-poky-rpi" instead of "meta-raspberrypi" (we already know the commit is meta-raspberrypi related as it's in this repository :) )

texierp:
> I like this idea, it will make testing a little easier. I'd recommend the following changes though:
> 
> * Put 'poky' in the name of the kas file so we know this uses the poky distro, maybe "kas-poky-rpi.yml".
> * Add the example of how to use this to the readme, maybe in a section "Quick start with kas" after the existing "Quick start" section.
> * Re-word the commit summary slightly to begin with "kas-poky-rpi" instead of "meta-raspberrypi" (we already know the commit is meta-raspberrypi related as it's in this repository :) )

Thanks @pbrkr for the review, very useful.

I will update the PR with your remarks :)

kas-poky-rpi.yml
```diff
@@ -0,0 +1,58 @@
...
+machine: raspberrypi4
+distro: poky
+target:
+  - core-image-base
+
+repos:
+  meta-raspberry:
+
+  poky:
+    url: https://git.yoctoproject.org/git/poky
```
kraj: It would be preferred if we can use oe-core/bitbake repos here.

texierp: @kraj In fact, I tried to create a kas file that follows the dependency recommendations in the README file (to facilitate the build configuration for the end user) with poky and so on. 

And as suggested by @pbrkr , I renamed this one kas-poky*.

So, maybe we can add another kas file for a "no-distro" configuration with the oe-core/bitbake repos ?

kraj: thats fine, I see that we ask for poky too but to keep in sync with readme nodistro is important too

texierp: Sure @kraj , I will add a kas file for nodistro in another PR ;)

agherzan:
This is interesting. I haven't heard of this tool before. I might even use it :) Is there any chance we can support includes?

texierp:
> This is interesting. I haven't heard of this tool before. I might even use it :) Is there any chance we can support includes?

Indeed nice tool ;).

With includes, you mean something like that https://gitlab.denx.de/Xenomai/xenomai-images ?

agherzan:
Exactly. That looks pretty nice.

texierp:
Sure, I will do that ;)

PR #564, opened 2020-01-17
rpi-base.inc: generate wic images by default

texierp:
Also enable wic.bmap image generation.

From the documentation in [1], Bmaptool is a generic tool
for creating the block map (bmap) for a file and copying
files using the block map. The idea is that large files,
like raw system image files, can be copied or flashed a
lot faster and more reliably with bmaptool than with
traditional tools, like "dd" or "cp".

Example:

$: sudo bmaptool copy 'image-name'.wic.xz /dev/'your-block-device'

[1] - https://github.com/intel/bmap-tools

Signed-off-by: Pierre-Jean Texier <pjtexier@koncepto.io>

### DIFF
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -5,7 +5,7 @@ include conf/machine/include/rpi-default-providers.inc
 SOC_FAMILY = "rpi"
 include conf/machine/include/soc-family.inc
 
-IMAGE_FSTYPES ?= "tar.bz2 ext3 rpi-sdimg"
+IMAGE_FSTYPES ?= "tar.bz2 ext3 wic.bz2 wic.bmap"
 WKS_FILE ?= "sdimage-raspberrypi.wks"
 
 XSERVER = " \


texierp:
Thanks @pbrkr for the review

> I think wic should be the default instead of rpi-sdimg these days. Could you remove rpi-sdimg from the list?
> 

Indeed, I will drop `rpi-sdimg` from the `IMAGE_FEATURES` list.

> xz gives a great compression ratio but at the cost of additional build time. I think bz2 may be a better default.

You're right, I will update accordingly.

texierp:
It seems we have to update the CI [config](https://github.com/agherzan/jenkins-conf-yocto-ci.resin.io/blob/master/jobs/meta-raspberrypi-generic/config.xml#L214) too, as it causes a build failure [1] in fact.

[1] - https://yocto-ci.resin.io/job/meta-raspberrypi-generic/1812/consoleText

agherzan:
Looking into it.

pbrkr:
@agherzan can you pin poky and meta-oe to older commits on master for CI for now and then unpin once the python2 fallout is handled?

agherzan:
I'll try to do that if this is not going to get fixed soon. I was hoping to not have to do it.

agherzan:
Merged as part of: https://github.com/agherzan/meta-raspberrypi/pull/569 . Thanks.

PR #568, opened 2020-01-20
RPI CONFIG: Disable UART explicitly, fixes #567

### LINKED ISSUE #567: "ENABLE_UART=0" do not disables console totally
**Description**

When setting ENABLE_UART=0 to disable the console on GPIO14/15, the kernel boot log is not shown, but the console is started and the login prompt is displayed. It is possible to login and use the terminal.

**Steps to reproduce the issue:**
1. Build with ENABLE_UART=0
2. Load the raspberry pi (I'm using CM3+)
3. Open serial console and boot the pi
4. At the end of the boot the prompt is displayed on the console

**Describe the results you received:**

The console is available on the UART

**Describe the results you expected:**

The console should not be available at all on the UART, instead the UART should remain available for other usage

**Additional information you deem important (e.g. issue happens only occasionally):**

I have solved the issue setting "enable_uart=0" in the "config.txt" file. When ENABLE_UART=0 is set, the config file is just not filled. Instead it should be completed with "enable_uart=0".

**Additional details (revisions used, host distro, etc.):**

I'm using warrior branch

I can propose à PR if you indicate me on which branch I should do it. master ? How to apply on all branches then ?

myfreescalewebpage:
**- What I did**

Explicitly set "enable_uart=0" when the console is disabled.

**- How I did it**

Modify rpi-config recipe.


recipes-bsp/bootfiles/rpi-config_git.bb
```diff
@@ -159,6 +159,9 @@ do_deploy() {
     if [ "${ENABLE_UART}" = "1" ]; then
         echo "# Enable UART" >>${DEPLOYDIR}/bcm2835-bootfiles/config.txt
         echo "enable_uart=1" >>${DEPLOYDIR}/bcm2835-bootfiles/config.txt
+    else
+        echo "# Disable UART" >>${DEPLOYDIR}/bcm2835-bootfiles/config.txt
+        echo "enable_uart=0" >>${DEPLOYDIR}/bcm2835-bootfiles/config.txt
```
agherzan: I'm wondering if setting an explicit disable would make more sense. This is because I think the current assumption is that when we don't set `ENABLE_UART` the value ends up being whatever the firmware uses by default. @kraj @pbrkr What do you guys think?

kraj: it should be that enable/disable knob overrides firmware defaults regardless

agherzan: I'm not sure I understand what you mean. All I'm saying is that with this changing we move from default whatever the firmware states per machine -> default force disable.

myfreescalewebpage: To propose an alternative, the value of "ENABLE_UART" can "1" to explicitly enable, "0" to explicitly disable, "" [empty] to keep default.

agherzan: That would work for me.

myfreescalewebpage: Anyone else ?

myfreescalewebpage:
Added explicit disable as discussed with @agherzan 
Also update doc (let me known if the text is ok for you).

agherzan:
@agherzan-ci ok to test

myfreescalewebpage:
Merged master in my branch to allow merge of the PR.

myfreescalewebpage:
@agherzan any news here ?

agherzan:
@my can you please squash the commits? We ended up with two commits and a merge one. Let's have only one on top of current HEAD.

myfreescalewebpage:
Sorry for the delay.
I'm not sure of the current commit... I'm not a git expert on this, maybe you can assist me because I don't know how and why doing this. Thanks!

agherzan:
`gIt rebase -i` and squash all the bits.

myfreescalewebpage:
Miss up with my git, deleted and created again this PR at https://github.com/agherzan/meta-raspberrypi/pull/588

PR #587, opened 2020-02-25
Kernel update

drewmoseley:
**- What I did**
Cherry picked kernel update commits from zeus.

On the Raspberry Pi 4, with the Pi Foundation Display, I needed these to get Xorg working with VC4 graphics.  Without these updates, Xorg launches and the screen displays the mouse cursor and responds to touch, but no clients are displayed.

**- How I did it**


drewmoseley:
Technically I didn't need the 4.19.93 update; the display started working with 4.19.88 but it seems worthwhile to move all the way.

Also, I cannot get this to work on RpI3.  I see log messages in /var/log/Xorg.0.log about modeset  checking for HDMI-1 and Composite-1 but no DSI.  Where on the RpI4 I do see DSI.  Any thoughts on how to address that?

agherzan:
Sorry for the late reply but I was thinking about this lately. The issue is that we do not actively maintain old branches and in the past, we agreed to only backport fixes and not provide components updates to avoid regressions. I'm a little bit reticent in pushing in updates that don't target specific issues. CC @pbrkr

drewmoseley:
Are others able to get X/Sato working without at least the 4.19.88 update?

agherzan:
I was more targetting this part of your comment "Technically I didn't need the 4.19.93 update".

drewmoseley:
Gotcha.  I'll wait for @pbrkr to comment but can easily drop that commit.

drewmoseley:
@pbrkr any thoughts here?

pbrkr:
@agherzan, @drewmoseley: I'm currently off work due to illness, sadly can't promise a reply to anything for probably another 2 weeks. Please go ahead without my input.

drewmoseley:
@pbrkr be well.  Sorry to hear but we await your return.

I'll just drop the extra patch here.

agherzan:
Let's merge it @drewmoseley . I'll blame youi if it breaks. :)

drewmoseley:
Hah.  Works for me.

agherzan:
Cheers @drewmoseley

PR #619, opened 2020-04-09
RFC: raspberrypi3: Switch to vc4-fkms-v3d.

drewmoseley:
This seems to be required to have a functional DSI display.

Signed-off-by: Drew Moseley <drew.moseley@northern.tech>

**- What I did**

Troubleshooting of blank Xorg display using the Pi Foundation display.

**- How I did it**

After changing this, core-image-sato works on both HDMI and the DSI display.

### DIFF
--- a/conf/machine/raspberrypi3-64.conf
+++ b/conf/machine/raspberrypi3-64.conf
@@ -31,4 +31,5 @@ KERNEL_IMAGETYPE_UBOOT ?= "Image"
 KERNEL_IMAGETYPE_DIRECT ?= "Image"
 KERNEL_BOOTCMD ?= "booti"
 
+VC4DTBO ?= "vc4-fkms-v3d"
 ARMSTUB ?= "armstub8.bin"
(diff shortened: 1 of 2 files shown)


drewmoseley:
I am not sure if this is right but I just started looking into the differences between the Pi4 and Pi3 setups since the Pi foundation display works properly on the Pi4 but not the Pi3 when running Xorg based images.

Please let me know if there is a better way to handle this.

kraj:
> I am not sure if this is right but I just started looking into the differences between the Pi4 and Pi3 setups since the Pi foundation display works properly on the Pi4 but not the Pi3 when running Xorg based images.
> 
> Please let me know if there is a better way to handle this.

if you are using mesa+vc4hgraphics then this perhaps is ok

drewmoseley:
Yeah, I'm just using the default config in warrior and newer.  Is this something that should be updated in the repo or just kept as a local change?

kraj:
> Yeah, I'm just using the default config in warrior and newer. Is this something that should be updated in the repo or just kept as a local change?
I think for master/dunfell it should be needed perhaps because there we default to vc4graphics for sure, older branches I don't know

PR #620, opened 2020-04-10
rpi-base.inc: Include modules if I2C is enaled

leon-anavi:
Add kernel modules i2c-dev and i2c-bcm2708 to variable
MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS if ENABLE_I2C is set to 1.
This affects images based on packagegroup-core-boot, including
the core-image-minimal image.

Signed-off-by: Leon Anavi <leon.anavi@konsulko.com>

**- What I did**

The patch ensures that kernel modules` i2c-dev` and `i2c-bcm2708` are included in `core-image-minimal` if ENABLE_I2C is set to 1. This fix is for all images based on `packagegroup-core-boot`, including the `core-image-minimal` image which otherwise doesn't include`kernel-modules` because they are set in variable `MACHINE_EXTRA_RRECOMMENDS` which does **not** have effect for `core-image-minimal`. The patch improves the developer experience while working on `core-image-minimal` with I2C support.

**- How I did it**

I added Linux kernel modules `i2c-dev` and `i2c-bcm2708` to variable `MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS` if I2C is enabled, aka if variable `ENABLE_I2C` is set to 1. This affects images based on `packagegroup-core-boot`, including the `core-image-minimal image`.

Best regards, Leon

### DIFF
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -76,7 +76,7 @@ MACHINE_FEATURES += "apm usbhost keyboard vfat ext2 screen touchscreen alsa blue
 # Raspberry Pi has no hardware clock
 MACHINE_FEATURES_BACKFILL_CONSIDERED = "rtc"
 
-MACHINE_EXTRA_RRECOMMENDS += " kernel-modules udev-rules-rpi"
+MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " kernel-modules udev-rules-rpi"
 
 # Set Raspberrypi splash image
 SPLASH = "psplash-raspberrypi"


conf/machine/include/rpi-base.inc
```diff
@@ -77,6 +77,7 @@ MACHINE_FEATURES += "apm usbhost keyboard vfat ext2 screen touchscreen alsa blue
 MACHINE_FEATURES_BACKFILL_CONSIDERED = "rtc"
 
 MACHINE_EXTRA_RRECOMMENDS += " kernel-modules udev-rules-rpi"
+MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " ${@oe.utils.conditional("ENABLE_I2C", "1", "kernel-module-i2c-dev kernel-module-i2c-bcm2708", "", d)}"
```
kraj, line 80: why dont we add it to `MACHINE_EXTRA_RRECOMMENDS` ?

leon-anavi: According to Yocto mega manual variable `MACHINE_EXTRA_RRECOMMENDS` affects only images based on `packagegroup-base`, which does **not** include the `core-image-minimal` or `core-image-full-cmdline` images. [For more details please have a look here](https://www.yoctoproject.org/docs/current/mega-manual/mega-manual.html#var-MACHINE_EXTRA_RRECOMMENDS).

Thanks, Leon

kraj: then lets convert the `MACHINE_EXTRA_RRECOMMENDS` to `MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS` while you are here.

leon-anavi: OK, sounds good to me. I will update the GitHub pull request. Just give me a moment. Thanks for the feedback!

shr-project:
Why was it changed to install all kernel-modules instead of selected subset?

In many builds kernel-modules contains *a lot* of modules most of which won't ever be used and significantly slow e.g. opkg upgrades, because kernel with all modules is upgraded quite often.

I'm already removing kernel-modules also from MACHINE_EXTRA_RRECOMMENDS so I'll just adjust it to remove it from the other variable as well, but would like to read the reasoning for including all in image.

agherzan:
Being a "recommends" variable it was a shortcut for people not have to cherry-pick the right modules for them. Semantically, the layer recommends the images to push in the kernel modules associated to the kernel image. Most of the users will just expect all the needed kernel modules in place. To be honest it smells a bit as an enforced policy but I guess it wins on the convenience side.

shr-project:
> people not have to cherry-pick the right modules

meta-raspberrypi could cherry-pick the right subset

but OK, it's not too difficult to replace with small list in own layer.

BTW: the leading space in both these assignments isn't needed as there is +=.

agherzan:
> > people not have to cherry-pick the right modules
> 
> meta-raspberrypi could cherry-pick the right subset

But we would enforce even more policy in this way. People will start asking about non-rpi specific kernel modules. I'll think about as I do agree this is not great. Also, we would need to keep track of the kernel modules in the rpi kernel. 

> BTW: the leading space in both these assignments isn't needed as there is +=.

Yup. It was a miss.

leon-anavi:
Hi @shr-project,

For the record, the first version of the patch was only including I2C related modules:
```
MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " ${@oe.utils.conditional("ENABLE_I2C", "1", "kernel-module-i2c-dev kernel-module-i2c-bcm2708", "", d)}"
```

Following a code review and a discussion with @kraj, due to the arguments stated by @agherzan I have modified the patch to its current state.

Best regards, Leon

kraj:
core-image-minimal, we do not expect that to have all kind of features, its just boots into a console and give your network, so adding logic to include modules into core-image-minimal is not something we should look for. `core-image-base` is more for what you are trying to do here.

kraj:
@agherzan I think @shr-project has a good point here, I think its easy to construct from small to big  then other way around, We should revert back to using `MACHINE_EXTRA_RRECOMMENDS`  since then we don't say its essential.

leon-anavi:
> We should revert back to using `MACHINE_EXTRA_RRECOMMENDS` since then we don't say its essential.

If you are going to revert it, can I submit again the initial version of the patch which as ensuring that I2C related modules (`kernel-module-i2c-dev` and `kernel-module-i2c-bcm2708`) are present in `core-image-minimal` when `ENABLE_I2C` is set to 1?

Without this fix, even when `ENABLE_I2C` is set to 1, I2C will still not work out of the box for `core-image-minimal`.

kraj:
I am not too enthusiastic about making core-image-minimal more functional than what it is. but since its enabled via a knob it perhaps is fine, my questions then would be, will it work for all cases ( all kind of images ) if not then see if it can be, core-image-minimal is just a reference image from oe-core and end users might be constructing their own images and if we are adding a knob it should work in all cases.

agherzan:
So, what shall we do? I agree with @leon-anavi 's proposal of only including the i2c modules based on  ENABLE_I2C.

kraj:
v1 of patch should be re-instated.

leon-anavi:
OK, I will take of it and I will submit a new GitHub pull request.

Thanks,
Leon

agherzan:
Thanks @leon-anavi

leon-anavi:
The new patch is available for a review as GitHub pull request: https://github.com/agherzan/meta-raspberrypi/pull/635

Thanks,
Leon

PR #624, opened 2020-04-13
psplash: Do not define ALTERNATIVE_PRIORITY for non-existing provider

kraj:
This simply causes build warnings about priority of two packages being
same, but infact this is redundant, therefore remove setting ALTERNATIVE_PRIORITY
for psplash-raspberrypi


agherzan:
I'm pretty sure we included this to fix the priority of picking the right boot logo of rpi. Isn't this going to affect it?

kraj:
> I'm pretty sure we included this to fix the priority of picking the right boot logo of rpi. Isn't this going to affect it?

@agherzan the logo file is added as C file and that is overridden for rpi from this layer this should appear before oe-core in BBPATHs. So it works well, priority won't help you if you have BBPATH order reversed

kraj:
@agherzan do you need more information on this?

agherzan:
All good. Merged.

bernardoaraujor:
I'm trying to overwrite the RPi splash screen.
Followed the instructions from https://stackoverflow.com/questions/45788396/yocto-raspberry-pi-change-psplash-image
no success.

any tips?

PR #636, opened 2020-04-29
merge master changes into dunfell

kraj:


agherzan:
@kraj this will need a rebase.

kraj:
> @kraj this will need a rebase.

actually this is a proper merge from master with one commit reverted which is not required on
dunfell, so you see the history but actually diff is is only two newest commits in master. This should help future merges as we perhaps will get some frequent patches for future.

I could have just cherry-picked two commits. Do you prefer that ?

agherzan:
Uh, I see. I'd definitely prefer cherry-picks.

agherzan:
Thanks @kraj

PR #649, opened 2020-05-26
Update runtime dependency from -imaging to -pillow

dl9pf:
bitbake world will fail with ERROR: Nothing RPROVIDES 'python3-imaging' . 

Replace with python3-pillow .


dl9pf:
https://en.wikipedia.org/wiki/Python_Imaging_Library 
>> Python Imaging Library (abbreviated as PIL) (in newer versions known as Pillow)   <<

kraj:
this is now addressed via 07fc972

PR #651, opened 2020-05-26
 conf/layer.conf: Add LAYERDEPENDS_raspberrypi 

dl9pf:
To allow the command
  bitbake-layers layerindex-fetch meta-raspberrypi

to work out, we need to declare the dependencies.
Use the same dependencies as declared in kas-poky-rpi.yaml .

Question is if we add the multimedia layer as well
(as mentioned in the README.md but not in the kas-poky-rpi.yaml).

 
LAYERDEPENDS_raspberrypi = "core networking-layer perl-layer qt5-layer"

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -10,6 +10,7 @@ BBFILE_PATTERN_raspberrypi := "^${LAYERDIR}/"
 BBFILE_PRIORITY_raspberrypi = "9"
 
 LAYERSERIES_COMPAT_raspberrypi = "sumo thud warrior zeus dunfell"
+LAYERDEPENDS_raspberrypi = "core meta-python"
 
 # Additional license directories.
 LICENSE_PATH += "${LAYERDIR}/files/custom-licenses"


conf/layer.conf
```diff
@@ -10,6 +10,7 @@ BBFILE_PATTERN_raspberrypi := "^${LAYERDIR}/"
 BBFILE_PRIORITY_raspberrypi = "9"
 
 LAYERSERIES_COMPAT_raspberrypi = "sumo thud warrior zeus dunfell"
+LAYERDEPENDS_raspberrypi = "core networking-layer perl-layer qt5-layer"
```
pbrkr, line 13: `networking-layer` and `qt5-layer` are handled by the `BBFILES_DYNAMIC` code below so those are optional dependencies.

kraj: we should only add required layers not ones under dynamic layers they are optional.

dl9pf: Fully agree on the point to trim this down. Ideally to 'core' only. 

I stumbled across this when dealing with the sense-hat which has a dependency to "meta-python" (which is implied by networking-layer in above patch). 

So if we trim it down, we end-up with
LAYERDEPENDS_raspberrypi = "core meta-python"

This will expand to  core + meta-python + openembedded-layer (meta-oe) .

Or do you prefer to move the python dependencies into a dynamic-layer and modify the packagegroups ? Affected would be from a quick look:
 - recipes-core/packagegroups/packagegroup-rpi-test.bb  has:
            * python3-rtimu
            * python3-sense-hat
 - recipes-devtools/python/*   has
            * python3-rtimu_git.bb
            * python3-sense-hat
            * rpi-gpio
            * rpio

We'd have to move/modify these if doing a dynamic layer inclusion for meta-python.

Thoughts ?

agherzan: I think the refactoring of meta-python could be treated separately and considered for now "core".

dl9pf:
Updated.

conf/layer.conf
kraj: Only recipe that appears to be dependent on meta-python is `python3-sense-hat` which rdeps on `python3-imaging` provided by `meta-python/recipes-devtools/python/python3-pillow` recipe, I think it will be good to investigate this aspect and perhaps remove this rdep via packageconfig or move pillow recipes to oe-core

agherzan: This would be nice. Locking people into using/pulling meta-python might not be great.

dl9pf: what about moving the sense hat into the dynamic layer https://github.com/agherzan/meta-raspberrypi/tree/master/dynamic-layers/meta-python .

dl9pf:
> Can you please edit your commit log following https://meta-raspberrypi.readthedocs.io/en/latest/contributing.html#formatting-patches? Otherwise, looks good to me.

Fixed.

conf/layer.conf
agherzan: I think this would work. @kraj ?

kraj: yes I think so too

agherzan: Let's do that @dl9pf and get this over the line.

agherzan: Small ping for this one @dl9pf .

kraj:
this is now addressed via 08d27cd and 25d1eb4

PR #660, opened 2020-06-20
u-boot: Fix booting raspberrypi CM3 module

nandra:
With stock u-bootv2020.01 which we use from poky CM3 module won't boot.
We need to apply this patch in order to have platform again working.

Add it only for cm3 to not break anythign else.

Signed-off-by: Marek Belisko <marek.belisko@open-nandra.com>

### DIFF
--- a/recipes-bsp/u-boot/u-boot/0001-dm-core-Move-ofdata_to_platdata-call-earlier.patch
+++ b/recipes-bsp/u-boot/u-boot/0001-dm-core-Move-ofdata_to_platdata-call-earlier.patch
@@ -1,0 +1,51 @@
+From 336d86ebd146905cf4384912f4f27699b6e37c72 Mon Sep 17 00:00:00 2001
+From: Simon Glass <sjg@chromium.org>
+Date: Sun, 29 Dec 2019 21:19:17 -0700
+Subject: [PATCH] dm: core: Move ofdata_to_platdata() call earlier
+
+This method is supposed to extract platform data from the device tree. It
+should be done before the device itself is probed. Move it earlier in the
+device_probe() function.
+
+Upstream-Status: Backport
+
+Signed-off-by: Simon Glass <sjg@chromium.org>
+---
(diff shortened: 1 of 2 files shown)


recipes-bsp/u-boot/u-boot/0001-dm-core-Move-ofdata_to_platdata-call-earlier.patch
```diff
@@ -0,0 +1,49 @@
+From 336d86ebd146905cf4384912f4f27699b6e37c72 Mon Sep 17 00:00:00 2001
+From: Simon Glass <sjg@chromium.org>
+Date: Sun, 29 Dec 2019 21:19:17 -0700
+Subject: [PATCH] dm: core: Move ofdata_to_platdata() call earlier
+
+This method is supposed to extract platform data from the device tree. It
+should be done before the device itself is probed. Move it earlier in the
+device_probe() function.
+
+Signed-off-by: Simon Glass <sjg@chromium.org>
```
agherzan, line 10: Can we also add an upstream status footer?

nandra: Patch is taken from latest u-boot tree so it's already in tree.

agherzan: Nice. Can we have that specified (backported)?

nandra:
@agherzan Added Upstream-Status to patch. Thanks.

agherzan:
@nandra Is this only required for dunfell?

nandra:
> Is this only required for dunfell?

Yes.

vkotarov:
Not sure if this is the right place to comment but the patch somehow made it to gatesgarth. On the other hand, it's not needed there as u-boot version is 2020.07 and already has it applied. Eventually, the do_patch task breaks on gatesgarth.

PR #676, opened 2020-07-08
linux-raspberrypi: bump to 5.4.50 since upstream was force-pushed

dozylynx:

Fixes the build since the upstream branch was force-pushed, so the
commit referenced in the SRCREV was no longer present in the branch.

Briefly boot-tested on a Raspberry Pi 4 with Xen via meta-virtualization.


agherzan:
Merged. Thanks.

PR #684, opened 2020-07-16
sdcard_image-rpi.bbclass: enable extensible inclusion into boot

dozylynx:
Add DEPLOYPAYLOAD, similar to the existing FATPAYLOAD, to enable
adding files to the boot partition from the image deploy directory.
Files such as hypervisor binaries may not be present (and in fact
unwanted) within the root filesystem, so FATPAYLOAD is not sufficient.

DEPLOYPAYLOAD is implemented with support for file renaming from the
source file in the image deploy directory to the filename written into
the boot image. DEPLOYPAYLOAD is a space-separated list of entries for
additions, each of which is colon-separated:
>     <image deploy directory file>:<destination filename>

This is used for including Xen, which produces a machine-specific
file in the image deploy directory, with its expected filename: xen.

Files that are to be included from the image deploy directory will
be produced by tasks that the do_image_rpi_sdimg[depends] must list,
so enable adding entries to that via a new variable:
RPI_SDIMG_EXTRA_DEPENDS.

These changes enable retiring a Xen-specific Raspberry Pi SD card
bbclass from meta-virtualization and have been tested on the
Raspberry Pi 4.

### DIFF
--- a/classes/sdcard_image-rpi.bbclass
+++ b/classes/sdcard_image-rpi.bbclass
@@ -59,6 +59,7 @@ do_image_rpi_sdimg[depends] = " \
     ${@bb.utils.contains('MACHINE_FEATURES', 'armstub', 'armstubs:do_deploy', '' ,d)} \
     ${@bb.utils.contains('RPI_USE_U_BOOT', '1', 'u-boot:do_deploy', '',d)} \
     ${@bb.utils.contains('RPI_USE_U_BOOT', '1', 'u-boot-default-script:do_deploy', '',d)} \
+    ${RPI_SDIMG_EXTRA_DEPENDS} \
 "
 
 do_image_rpi_sdimg[recrdeps] = "do_build"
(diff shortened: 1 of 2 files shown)


dozylynx:
Thanks for the review and encouragement. I've just added a second commit with an update to the docs to describe the new variables that this PR introduces, plus the existing FATPAYLOAD variable.

classes/sdcard_image-rpi.bbclass
```diff
@@ -148,6 +149,17 @@ IMAGE_CMD_rpi-sdimg () {
         fi
     fi
 
+    # Add files (eg. hypervisor binaries) from the deploy dir
+    if [ -n "${DEPLOYPAYLOAD}" ] ; then
+        echo "Copying deploy file payload into VFAT"
+        for entry in ${DEPLOYPAYLOAD} ; do
+            # Split the entry at ':' to enable file renaming for the destination
+            DEPLOY_FILE="$(echo $entry | cut -f1 -d:)"
+            DEST_FILENAME="$(echo $entry | cut -f2 -d:)"
```
agherzan, line 158: Would make sense to default DEST_FILENAME to DEST_FILENAME when not provided? So functionally `example2:example2` would be the same as `example2`.

dozylynx: ack - have just refreshed the branch with that implemented and documented

PR #685, opened 2020-07-16
rpi-config: Add CAN_OSCILLATOR variable to set mcp2515 crystal frequency

Limoto:
Add CAN_OSCILLATOR variable to make it possible to set frequency of the crystal oscillator used on the actual CAN board.

Current configuration (16MHz) is made for the PiCAN2 board, that uses 16MHz crystal. I use Waveshare RS485 CAN HAT that has 8MHz crystal soldered (although according to Waveshare there is also a 16MHz crystal version of the board).

### DIFF
--- a/recipes-bsp/bootfiles/rpi-config_git.bb
+++ b/recipes-bsp/bootfiles/rpi-config_git.bb
@@ -28,6 +28,8 @@ VC4DTBO ?= "vc4-kms-v3d"
 GPIO_IR ?= "18"
 GPIO_IR_TX ?= "17"
 
+CAN_OSCILLATOR ?= "16000000"
+
 inherit deploy nopackages
 
 do_deploy() {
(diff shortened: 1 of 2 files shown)


Limoto:
* Added Signed-off-by 
* Added documentaton update

Limoto:
Rebased. It got a bit confusing now with the Dual CAN support, however, as long as there is no common board with Dual CAN and other crystal than 16MHz, it should be fine.

recipes-bsp/bootfiles/rpi-config_git.bb
```diff
@@ -206,7 +208,7 @@ do_deploy() {
     # ENABLE CAN
     elif [ "${ENABLE_CAN}" = "1" ]; then
```
agherzan, line 209: I reckon this should now be also used in `ENABLE_DUAL_CAN`. @colin-pm , what do you think?

colin-pm: I think for consistency the variable might as well get used for single and dual CAN interfaces so this wont have to be revisited when a dual CAN board shows up on the market with a different oscillator.

agherzan: Agreed. Let's do that and move the variable description in docs after `ENABLE_DUAL_CAN`.

Limoto: Alright, I added the configruation also for dual CAN and combined the documentation together.

PR #694, opened 2020-08-05
u-boot: Move fw_env.config to u-boot append

fbertux:
The u-boot-env is provided by u-boot recipe and not by libubootenv, so
right recipe to append is the u-boot.

Adding the rpi-u-boot-scr in DEPENDS variable is wrong because it is
forcing rpi-u-boot-scr to be a dependency, but it'll fail if I have
another recipe that provides bootscript, once both recipes provide the
same file. The default value of u-boot-default-script is rpi-u-boot-scr,
so right way is to use u-boot-default-script and change the
PREFERRED_PROVIDER_u-boot-default-script if needed.

Signed-off-by: Fabio Berton <fabio.berton@ossystems.com.br>
(cherry picked from commit fc72c061411ea78515a670b63b0d44c13451cb78)


agherzan:
Thanks. Merged.

Issue #696, opened 2020-08-06
u-boot fetch error for fw_env.config in dunfell


mtrensch:
**Description**

Fetching u-boot fails as "fw_env.config" cannot be found:
```
ERROR: u-boot-1_2020.01-r0 do_fetch: Fetcher failure for URL: 'file://fw_env.config'. Unable to fetch URL from any source.
ERROR: Logfile of failure stored in: /build/build_raspberrypi/tmp/work/niot_e_tpi51_en_re-poky-linux-gnueabi/u-boot/1_2020.01-r0/temp/log.do_fetch.25975
NOTE: recipe u-boot-1_2020.01-r0: task do_fetch: Failed
ERROR: Task (/build/build_raspberrypi/../poky/meta/recipes-bsp/u-boot/u-boot_2020.01.bb:do_fetch) failed with exit code '1'
```

**Steps to reproduce the issue:**
1. build u-boot (bitbake u-boot)

**Describe the results you received:**

```
ERROR: u-boot-1_2020.01-r0 do_fetch: Fetcher failure for URL: 'file://fw_env.config'. Unable to fetch URL from any source.
ERROR: Logfile of failure stored in: /build/build_raspberrypi/tmp/work/niot_e_tpi51_en_re-poky-linux-gnueabi/u-boot/1_2020.01-r0/temp/log.do_fetch.25975
NOTE: recipe u-boot-1_2020.01-r0: task do_fetch: Failed
ERROR: Task (/build/build_raspberrypi/../poky/meta/recipes-bsp/u-boot/u-boot_2020.01.bb:do_fetch) failed with exit code '1'
```

**Describe the results you expected:**
Building works.

**Additional information you deem important (e.g. issue happens only occasionally):**

This bug was introduced by following commit:
https://github.com/agherzan/meta-raspberrypi/commit/fc494a35b513b05860c5a813239274017f1ef934

The file fw_env.config resides in "${THISDIR}/files" which is not in FILESEXTRAPATHS. Either move it to u-boot or add "${THISDIR}/files" to FILESEXTRAPATHS.

agherzan:
Should be fixed now.

PR #698, opened 2020-08-06
sdcard_image-rpi.bbclass: Fix when RPI_SDIMG_EXTRA_DEPENDS not defined

agherzan:
This includes a fix for https://github.com/agherzan/meta-raspberrypi/pull/684/files
CC @dozylynx

PR #699, opened 2020-08-06
Clean dunfell and bring master changes

agherzan:
I reverted the custom backports and cleanly added them in the master order so we can maintain the changes easier. Also, one specific change was reverted without backporting because it was broken (https://github.com/agherzan/meta-raspberrypi/pull/660 CC @nandra). This PR misses a `FILESEXTRAPATHS` so please re-push.

This also fixes https://github.com/agherzan/meta-raspberrypi/issues/696.

The only additional fix is `sdcard_image-rpi.bbclass: Fix when RPI_SDIMG_EXTRA_DEPENDS not defined`. That fix is also pushed against `master` (https://github.com/agherzan/meta-raspberrypi/pull/698).


mirzak:
What is the reason of reverting https://github.com/agherzan/meta-raspberrypi/pull/699/commits/775481dae292df6d17f81790e67307f312534216. Reason I ask is that it is a build breaker, when it was introduce it broke some of our builds and now it will break again if you revert it :)

agherzan:
@mirzak It actually adds it afterwards. This is because by doing it in the same order we avoid conflicts (in this case on `rpi-config_git.bb`) with other commits. The initial backport was not a clean cherry-pick.

shr-project:
LGTM and thanks for fixing that u-boot build.

agherzan:
Thanks for the reviews. Merged.

PR #712, opened 2020-08-28
linux-raspberrypi-5.4: backport a fix for perf build with -fno-common…

shr-project:
… from gcc-10

### DIFF
--- a/recipes-kernel/linux/files/0001-perf-cs-etm-Move-definition-of-traceid_list-global-v.patch
+++ b/recipes-kernel/linux/files/0001-perf-cs-etm-Move-definition-of-traceid_list-global-v.patch
@@ -1,0 +1,68 @@
+From 168200b6d6ea0cb5765943ec5da5b8149701f36a Mon Sep 17 00:00:00 2001
+From: Leo Yan <leo.yan@linaro.org>
+Date: Tue, 5 May 2020 21:36:42 +0800
+Subject: [PATCH] perf cs-etm: Move definition of 'traceid_list' global
+ variable from header file
+
+The variable 'traceid_list' is defined in the header file cs-etm.h,
+if multiple C files include cs-etm.h the compiler might complaint for
+multiple definition of 'traceid_list'.
+
+To fix multiple definition error, move the definition of 'traceid_list'
+into cs-etm.c.
+
+Upstream-Status: Backport [v5.8-rc1]
+
+Fixes: cd8bfd8c973e ("perf tools: Add processing of coresight metadata")
+Reported-by: Thomas Backlund <tmb@mageia.org>
+Signed-off-by: Leo Yan <leo.yan@linaro.org>
+Reviewed-by: Mathieu Poirier <mathieu.poirier@linaro.org>
+Reviewed-by: Mike Leach <mike.leach@linaro.org>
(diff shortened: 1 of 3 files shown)


kraj (dismissed):
lgtm, does it work on top of my pull for kernel upgrade

shr-project:
> lgtm, does it work on top of my pull for kernel upgrade

on top of already merged https://github.com/agherzan/meta-raspberrypi/pull/707 ?

yes, I was testing this against current master

kraj:
> > lgtm, does it work on top of my pull for kernel upgrade
> 
> on top of already merged #707 ?
> 
> yes, I was testing this against current master

ok, I was wondering if we will need them on dunfell too?

shr-project:
> ok, I was wondering if we will need them on dunfell too?

I don't think so as it's needed only with:
https://git.openembedded.org/openembedded-core/commit/?id=4ee733e1551d3960a23a600eb71a01cf7a51fa06
which probably shouldn't be backported to dunfell

kraj:
> > ok, I was wondering if we will need them on dunfell too?
> 
> I don't think so as it's needed only with:
> https://git.openembedded.org/openembedded-core/commit/?id=4ee733e1551d3960a23a600eb71a01cf7a51fa06
> which probably shouldn't be backported to dunfell
> 
> But it does no harm, so we can backport it to dunfell with the next kernel upgrade, just to keep the kernel recipes identical.

yeah if it does not harm then perhaps it should be, for sake of mainainance

agherzan (dismissed):

recipes-kernel/linux/files/0001-perf-cs-etm-Move-definition-of-traceid_list-global-v.patch
```diff
@@ -0,0 +1,66 @@
...
+Cc: Alexander Shishkin <alexander.shishkin@linux.intel.com>
+Cc: Jiri Olsa <jolsa@redhat.com>
+Cc: Mark Rutland <mark.rutland@arm.com>
+Cc: Namhyung Kim <namhyung@kernel.org>
+Cc: Peter Zijlstra <peterz@infradead.org>
+Cc: Suzuki Poulouse <suzuki.poulose@arm.com>
+Cc: Tor Jeremiassen <tor@ti.com>
+Cc: linux-arm-kernel@lists.infradead.org
+Link: http://lore.kernel.org/lkml/20200505133642.4756-1-leo.yan@linaro.org
+Signed-off-by: Arnaldo Carvalho de Melo <acme@redhat.com>
```
agherzan, line 30: is it worth it to have a small upstream status? Is it to be pushed to upstream?

shr-project: It's backport from upstream (since v5.8-rc1). I've checked latest rpi-5.4.y branch in linux-raspberrypi (currently 5.4.61 still doesn't have it).

agherzan: It would be nice to have an upstream status backport from 5.8 (to know to keep an eye on it). Even though on a second thought we will keep an eye on it anyway so I don't really have a strong feeling about it.

agherzan (approved):
Thanks!

PR #719, opened 2020-09-08
libcamera: Added a recipe file for libcamera

madhavank27:
libcamera middleware has supported many pipeline for frame capturing.
To enable the frame capture from raspberrypi board, added new
bbappend file, which will enable the raspberrypi pipeline while compiling.

Signed-off-by: Madhavan Krishnan <madhavan.krishnan@linaro.org>

**- What I did**

**- How I did it**

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -31,6 +31,8 @@ BBFILES_DYNAMIC += " \
     networking-layer:${LAYERDIR}/dynamic-layers/networking-layer/*/*/*.bbappend \
     qt5-layer:${LAYERDIR}/dynamic-layers/qt5-layer/*/*/*.bb \
     qt5-layer:${LAYERDIR}/dynamic-layers/qt5-layer/*/*/*.bbappend \
+    multimedia-layer:${LAYERDIR}/dynamic-layers/multimedia-layer/*/*/*.bb \
+    multimedia-layer:${LAYERDIR}/dynamic-layers/multimedia-layer/*/*/*.bbappend \
 "
 
 DEFAULT_TEST_SUITES_remove_rpi = "parselogs"
--- a/dynamic-layers/multimedia-layer/recipes-multimedia/libcamera/libcamera.bbappend
+++ b/dynamic-layers/multimedia-layer/recipes-multimedia/libcamera/libcamera.bbappend
@@ -1,0 +1,2 @@
+PACKAGECONFIG[raspberrypi] = "-Dpipelines=raspberrypi"
+PACKAGECONFIG_append_rpi = " raspberrypi"


recipes-multimedia/libcamera/libcamera.bbappend
```diff
@@ -0,0 +1,5 @@
+SUMMARY = "Linux libcamera framework"
```
agherzan, line 1: Any reason to redefine this? (http://cgit.openembedded.org/meta-openembedded/tree/meta-multimedia/recipes-multimedia/libcamera/libcamera.bb?h=master#n1)

recipes-multimedia/libcamera/libcamera.bbappend
```diff
@@ -0,0 +1,5 @@
...
+
+PACKAGECONFIG[raspberrypi] = " -Dpipelines=raspberrypi, ,"
+PACKAGECONFIG_append_rpi = " raspberrypi"
+
```
agherzan, line 5: Cleanup extra newlines please.

recipes-multimedia/libcamera/libcamera.bbappend
```diff
@@ -0,0 +1,5 @@
+SUMMARY = "Linux libcamera framework"
+
+PACKAGECONFIG[raspberrypi] = " -Dpipelines=raspberrypi, ,"
```
agherzan, line 3: There are some extra whitespaces you don't really need.

agherzan (changes requested):
Besides the comments I had in-line, please reward the git commit subject line to provide some meaningful information about the change. Also, we had some discussions in the past to remove a hard dependency on meta-multimedia cc @kraj and this commit now brings back that dependency. I would recommend using the dynamic layer facility to avoid this.

recipes-multimedia/libcamera/libcamera.bbappend
madhavank27: FYI @agherzan 

I have updated the recipe file, as per your comments.

kraj (commented):
I think subject line should be
` libcamera: Define packageconfig to enable rpi pipeline`

recipes-multimedia/libcamera/libcamera.bbappend
agherzan: Thanks. The remaining comment is abouve the subject line and dynamic layers.

agherzan: You can clean up the extra space between the commas too. If I'm not mistaken you can also omit the commas - I wouldn't bet on it though so do try it out.

madhavank27:
FYI @kraj and @agherzan 

Thanks for the comments, and I have updated the recipe.

agherzan:
What do you think about dynamic layers?

madhavank27:
FYI @agherzan 

Sorry for inconvenience. yesterday I didn't notice the following comments.

"Besides the comments I had in-line, please reward the git commit subject line to provide some meaningful information about the change. Also, we had some discussions in the past to remove a hard dependency on meta-multimedia cc @kraj and this commit now brings back that dependency. I would recommend using the dynamic layer facility to avoid this."

I have updated the pull request with dynamic layers.

madhavank27:
FYI @agherzan 

Shall you please verify this patch changes.

agherzan (approved):
This looks good now  @madhavank27 .

agherzan:
@kraj any comments (you were part of the review)?

kraj (approved):
LGTM

PR #721, opened 2020-09-12
bump to Linux version 5.4.64

texierp:
Tested on `rpi3`:

```
5896120 bytes read in 247 ms (22.8 MiB/s)
Saving Environment to FAT... OK
## Booting kernel from Legacy Image at 00080000 ...
   Image Name:   Linux-5.4.64-v7
   Image Type:   ARM Linux Kernel Image (uncompressed)
   Data Size:    5896056 Bytes = 5.6 MiB
   Load Address: 00008000
   Entry Point:  00008000
   Verifying Checksum ... OK
## Flattened Device Tree blob at 2eff8e00
   Booting using the fdt blob at 0x2eff8e00
   Loading Kernel Image
   Using Device Tree in place at 2eff8e00, end 2f002f5b

Starting kernel ...

[    0.000000] Booting Linux on physical CPU 0x0
[    0.000000] Linux version 5.4.64-v7 (oe-user@oe-host) (gcc version 10.2.0 (GCC)) #1 SMP Fri Sep 11 12:57:30 UTC 2020
[    0.000000] CPU: ARMv7 Processor [410fd034] revision 4 (ARMv7), cr=10c5383d
[    0.000000] CPU: div instructions available: patching division code
```

PR #722, opened 2020-09-14
linux-raspberrypi-5.4: revert 1 commit from upstream to fix lttng-mod…

shr-project:
…ules build

* fixes:
  lttng-modules/2.12.2-r0/lttng-modules-2.12.2/probes/../probes/lttng-tracepoint-event-impl.h:131:6: error: conflicting types for 'trace_writeback_queue_io'

* fixes:
  https://github.com/agherzan/meta-raspberrypi/pull/721


agherzan (dismissed):
Nice one. Have you given it a spin?

shr-project:
> Nice one. Have you given it a spin?

Yes, it fixes the build for me.

The issue is that writeback_queue_io was updated in lttng-modules in:
https://github.com/lttng/lttng-modules/commit/817e02f9e256b8f2cbb9e3125eb061757b83bd4a
which isn't included in 2.12.2 version currently in oe-core and even if it is, then it checks for kernel version 5.9 and won't work with this commit backported from 5.9 to rpi-5.4.y.

I'll update the commit with this info, so that it isn't only in the PR.

agherzan:
Cheers @shr-project

agherzan (approved):

Issue #732, opened 2020-09-28
Rpi 4 doesn't boot with u-boot when enable_uart=0 in config.txt


tolszak:
Regular dunfell branch with local.conf

```
MACHINE = "raspberrypi4-64"
RPI_USE_U_BOOT=1
ENABLE_UART=0
```
core-image-minimal doesn't boot. It looks like uboot image hangs for some reason. Sometimes rainbow splash screen blinks.

I tried updated u-boot from July 2020 - no luck.
I removed serial from rpi.h ENV_DEVICE_SETTINGS in u-boot - no luck.
Following patch fixes it for me: #733

Spark1114:
I have kind of the same issue on the CM3+ with the following config: 

```
MACHINE = "raspberrypi3-64"
ENABLE_UART = "1"
RPI_USE_U_BOOT = "1"
```

And these layers: 

```
poky/meta-openembedded/meta-python
poky/meta-openembedded/meta-networking
poky/meta-openembedded/meta-oe
poky/meta
poky/meta-poky
poky/meta-yocto-bsp
poky/meta-raspberrypi
```

Here is the console output: 
                                        
```
U-Boot 2020.01 (Jan 06 2020 - 20:56:31 +0000)
                                        
DRAM:  948 MiB                          
RPI Compute Module 3+ (0xa02100)
```



I am able to boot the system when using MACHINE = "raspberrypi-cm3" 

Let me know if you need more information.

cveilleux:
Confirming we are seeing the same issue on raspberrypi4-64 with `RPI_USE_U_BOOT = "1"`

We have not set ENABLE_UART. I believe it is related to RPI_USE_U_BOOT that causes the boot failure.

HDMI screen plugged. Rainbow flashes quickly, then screen remains black.

agherzan:
Duplicates https://github.com/agherzan/meta-raspberrypi/pull/733

EasyIP2023:
[...]

PR #733, opened 2020-09-29
Remove UART usage in U-Boot when ENABLE_UART=0

tolszak:
This fixes booting on Rasbpberry Pi 4 when config.txt doesn't
contain enable_uart=1. Seems like u-boot tries to write to serial
but serial doesn't exist and it stalls.
Related to issue: #732

### DIFF
--- a/recipes-bsp/bootfiles/rpi-config_git.bb
+++ b/recipes-bsp/bootfiles/rpi-config_git.bb
@@ -160,7 +160,7 @@ do_deploy() {
     fi
 
     # UART support
-    if [ "${ENABLE_UART}" = "1" ]; then
+    if [ "${ENABLE_UART}" = "1" -o "${RPI_USE_U_BOOT}" = "1" ]; then
         echo "# Enable UART" >>${DEPLOYDIR}/${BOOTFILES_DIR_NAME}/config.txt
         echo "enable_uart=1" >>${DEPLOYDIR}/${BOOTFILES_DIR_NAME}/config.txt
     fi


myfreescalewebpage:
@tolszak for yout information I previously worked on this because uart is used for linux console too and this should be explicitly disabled to free the uart interface for another usage: https://github.com/agherzan/meta-raspberrypi/pull/588

tolszak:
My issue is related to uboot. With enable_uart=0 it doesn't boot. So it seems not really related to your issue.

myfreescalewebpage:
Not directly you're right :)

agherzan (changes requested):
This needs more details on what actually is going on. It's also worth raising and pushing the patch to upstream. If they merge it there we can backport it here until uboot gets bumped. Also, mind the contributions conventions (git subject, description, etc) and try to keep the same tab/indentation in files. Don't forget Signed-off-by and Upstream-status.

On top of these, a change needs to first go in master (meta-raspberrypi), and once that lands, we can raise additional PRs for other branches (eg. dunfell).

tolszak:
I uploaded new fix, the old one seems not working anymore.
In short I think the issue is with u-boot kernel or device tree, it seems to expect serial port to be available while enable_uart = 0 in config.txt disables it. It causes device to hang just after loading u-boot kernel. So the patch I uploaded enforces enable_uart=1 in config.txt when RPI_USE_U_BOOT=1.

agherzan:
@tolszak do we know why this is the case? Does it sound like a bug in uboot? Have we checked with upstream? Also, if we are going for a workaround (which we will if we have details on the issue at hand) this fix needs to land first in `master`.

tolszak:
> @tolszak do we know why this is the case? Does it sound like a bug in uboot? Have we checked with upstream? Also, if we are going for a workaround (which we will if we have details on the issue at hand) this fix needs to land first in `master`.

I'm pretty sure that this bug is in u-boot. Unfortunately I don't have enough time currently to research and fix that (and I don't think it will change next weeks/months). After making a few assumptions, I just came to this workaround and left it, here, for community, in case someone else hits the same issue.

agherzan:
Understandable. Thanks for this.

hommeabeil:
I don't think it is related to u-boot well not directly and since I am not an expert in this subject it is hard for me to start debugging it. But if I use the `bootfiles` from the Zeus branch with the kernel and the the U-boot from the Dunfell branch I can get farther in the boot  sequence. In my case, I am not sure why, but the kernel seems to hang when trying to get the SD  card. Maybe the firmware does not properly configure something  .... And all of this append with `enable_uart=0`, hope this help some one. I will try to keep looking.

agherzan:
Closing due to inactivity. We can reopen if there is some interest on it.

PR #736, opened 2020-10-02
[warrior] linux-raspberrypi: Only deploy cmdline.txt for the main kernel

alexgg:
When multiple kernels are being built, not all of them can deploy the same file. A use case for this is when Linux is being used as a bootloader to kexec a new kernel, and the initial one is configured differently to be much smaller in size.

This patch assumes the same command  line arguments is used by all kernels.

This is tested on the warrior branch so I PR it there. Once approved I will open a PR for master.

### DIFF
--- a/recipes-kernel/linux/linux-raspberrypi.inc
+++ b/recipes-kernel/linux/linux-raspberrypi.inc
@@ -130,11 +130,13 @@ do_compile_append() {
 }
 
 do_deploy_append() {
-    # Deploy cmdline.txt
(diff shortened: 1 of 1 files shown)


kraj (approved):
lgtm

agherzan (approved):

agherzan:
@alexgg thanks for this. We would need to first include this in master and afterward backport it on other branches. Could you first raise this against master?

agherzan (changes requested):
Needs to be on master first,

agherzan (approved):
Looks good. Merged.

PR #746, opened 2020-10-12
linux-raspberrypi: Only deploy cmdline.txt for the main kernel

alexgg:
When multiple kernels are being built, not all of them can deploy the
same file.

This patch assumes the same command line arguments is used by all kernels.

Related to https://github.com/agherzan/meta-raspberrypi/pull/736


agherzan (approved):
Looks good. Merged.

PR #747, opened 2020-10-15
dunfell: Fix raspberrypi-cm3 boot when u-boot is used

nandra:


agherzan (changes requested):
Is this dunfell specific? Most probably not and if so, can you please raise it against master first?

nandra:
@agherzan pls see #748

agherzan:
It will come in through https://github.com/agherzan/meta-raspberrypi/pull/754 (your commit was cherrypicked there).

PR #750, opened 2020-10-19
libcamera: Enabled camera overlays define based

madhavank27:
Added imx219.dtbo file in the overlays, and added dtoverlay in
/boot/config.txt define based.

To enable SOny Raspberry pi camera imx219 sensor, need to set
the variable ENABLE_CAMERA_OVERLAY as 1.

In future, we can add any camera overlays here to support and
use them with libcamera.

Signed-off-by: Madhavan Krishnan <madhavan.krishnan@linaro.org>

**- What I did**

**- How I did it**

### DIFF
--- a/recipes-bsp/bootfiles/rpi-config_git.bb
+++ b/recipes-bsp/bootfiles/rpi-config_git.bb
@@ -178,6 +178,12 @@ do_deploy() {
         echo "dtoverlay=${VC4DTBO}" >> ${DEPLOYDIR}/${BOOTFILES_DIR_NAME}/config.txt
     fi
 
+    # Choose Camera Sensor to be used, default imx219 sensor
+    if [ "${RASPBERRYPI_CAMERA_V2}" = "1" ]; then
+        echo "# Enable Sony RaspberryPi Camera" >> ${DEPLOYDIR}/${BOOTFILES_DIR_NAME}/config.txt
+        echo "dtoverlay=imx219" >> ${DEPLOYDIR}/${BOOTFILES_DIR_NAME}/config.txt
+    fi
+
     # Waveshare "C" 1024x600 7" Rev2.1 IPS capacitive touch (http://www.waveshare.com/7inch-HDMI-LCD-C.htm)
     if [ "${WAVESHARE_1024X600_C_2_1}" = "1" ]; then
         echo "# Waveshare \"C\" 1024x600 7\" Rev2.1 IPS capacitive touch screen" >> ${DEPLOYDIR}/${BOOTFILES_DIR_NAME}/config.txt
--- a/recipes-kernel/linux/linux-raspberrypi.inc
+++ b/recipes-kernel/linux/linux-raspberrypi.inc
@@ -27,7 +27,8 @@ LINUX_VERSION_EXTENSION ?= ""
 
 # CMDLINE for raspberrypi
 SERIAL = "${@oe.utils.conditional("ENABLE_UART", "1", "console=serial0,115200", "", d)}"
-CMDLINE ?= "dwc_otg.lpm_enable=0 ${SERIAL} root=/dev/mmcblk0p2 rootfstype=ext4 rootwait"
+CMA ?= "${@oe.utils.conditional("RASPBERRYPI_CAMERA_V2", "1", "cma=64M", "", d)}"
+CMDLINE ?= "dwc_otg.lpm_enable=0 ${SERIAL} root=/dev/mmcblk0p2 rootfstype=ext4 rootwait ${CMA}"
 
 # Add the kernel debugger over console kernel command line option if enabled
 CMDLINE_append = ' ${@oe.utils.conditional("ENABLE_KGDB", "1", "kgdboc=serial0,115200", "", d)}'
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -42,6 +42,7 @@ RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
     overlays/w1-gpio.dtbo \
     overlays/gpio-ir.dtbo \
     overlays/gpio-ir-tx.dtbo \
+    overlays/imx219.dtbo \
     "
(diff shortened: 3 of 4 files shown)


recipes-bsp/bootfiles/rpi-config_git.bb
```diff
@@ -178,6 +178,13 @@ do_deploy() {
         echo "dtoverlay=${VC4DTBO}" >> ${DEPLOYDIR}/${BOOTFILES_DIR_NAME}/config.txt
     fi
 
+    # Choose Camera Sensor to be used, default imx219 sensor
+    if [ "${RASPBERRYPI_CAMERA_V2}" = "1" ]; then
```
agherzan, line 182: Please document this variable in the docs. See other examples there.

recipes-bsp/bootfiles/rpi-config_git.bb
```diff
@@ -178,6 +178,13 @@ do_deploy() {
...
+        echo "# Enable Sony RaspberryPi Camera" >> ${DEPLOYDIR}/${BOOTFILES_DIR_NAME}/config.txt
+        echo "dtoverlay=imx219" >> ${DEPLOYDIR}/${BOOTFILES_DIR_NAME}/config.txt
+        echo "core_freq_min=250" >> ${DEPLOYDIR}/${BOOTFILES_DIR_NAME}/config.txt
```
agherzan, line 185: Why do we need this?

recipes-kernel/linux/linux-raspberrypi.inc
```diff
@@ -27,7 +27,8 @@ LINUX_VERSION_EXTENSION ?= ""
 
 # CMDLINE for raspberrypi
 SERIAL = "${@oe.utils.conditional("ENABLE_UART", "1", "console=serial0,115200", "", d)}"
-CMDLINE ?= "dwc_otg.lpm_enable=0 ${SERIAL} root=/dev/mmcblk0p2 rootfstype=ext4 rootwait"
+CMA = "${@oe.utils.conditional("RASPBERRYPI_CAMERA_V2", "1", "cma=64M", "", d)}"
```
agherzan, line 30: Why do we need this?

agherzan (changes requested):

recipes-bsp/bootfiles/rpi-config_git.bb
madhavank27: Since libcamera opensource is using the bcm2835-unicam device to access the camera, we need to explicitly set the device tree overlay in the config,txt.

Then only, libcamera could able to acquire and access the unicam device RaspberrypI cAMERA v2.

Please have a look into following libcamera installation docs for your reference.

https://github.com/raspberrypi/documentation/blob/master/linux/software/libcamera/README.md

recipes-kernel/linux/linux-raspberrypi.inc
madhavank27: For camera device, minimum of 64M contiguous memory allocation is needed.
So, we have added this in the cmdline.txt.

recipes-bsp/bootfiles/rpi-config_git.bb
madhavank27: Ok, thanks.

madhavank27:
FYI @agherzan

recipes-bsp/bootfiles/rpi-config_git.bb
agherzan: I was asking about `core_freq_min`. I see in the readme you pointed out that it is a workaround. Do we know if it was fixed? Can we can it out?

> Currently we also need to update the GPU's core_freq_min though this will become unnecessary in due course after further updates.

madhavank27: Yes, in the latest driver, we no need this core_freq_min, and we had removed this now.

recipes-kernel/linux/linux-raspberrypi.inc
madhavank27: We had tried keeping the cma as 32M, and libcamera could not able to capture the HD resolution 1280x720.

We are getting the following error.
ERROR V4L2 v4l2_videodevice.cpp:1129 /dev/video0[cap]: Not enough buffers provided by V4L2VideoDevice
ERROR RPI raspberrypi.cpp:781 Failed to allocate buffers

When we try to capture in HD resolution 1280x720, libcamera is selecting the best resolution as 1920x1080.
And since libcamera is capturing the frames in 10 bit bayer format following memory is required by it.

1920x1080x10/8 = 2592000 bytes.

So libcamera is requesting for 4 buffers with above mentioned size.

So we considered, minimum of 64M for CMA is required to work with libcamera for HD streams.

madhavank27:
FYI @agherzan

recipes-kernel/linux/linux-raspberrypi.inc
kbingham: For reference, I believe when the firmware/GPU based camera drivers are used, they reserve 128MB of RAM (gpu_mem=128m)

kraj: Thanks @kbingham 
@agherzan  this is a new knob and is only activated when using `RASPBERRYPI_CAMERA_V2` I think we should be fine with this change

recipes-bsp/bootfiles/rpi-config_git.bb
agherzan: Good stuff!

recipes-kernel/linux/linux-raspberrypi.inc
agherzan: It's not really a new knob - it's a definition based on another "knob" :) But overall I'm fine with this. I was thinking if we could make it a bit more elegant. Let the builds crunch it and I'll merge afterwards.

agherzan:
@agherzan-ci ok to test

agherzan (approved):

PR #767, opened 2020-11-25
Fix need for meta-python and BSP recipe upgrades

kraj:


recipes-core/packagegroups/packagegroup-rpi-test.bb
```diff
@@ -17,13 +17,16 @@ RDEPENDS_${PN} = "\
...
-    python3-sense-hat \
     connman \
     connman-client \
     wireless-regdb-static \
     bluez5 \
 "
 
+RDEPENDS_${PN} += "\
+    ${@bb.utils.contains("BBFILE_COLLECTIONS", "meta-python", "python3-sense-hat", "", d)} \
+"
```
pbrkr, line 28: Could this be done via a `packagegroup-rpi-test.bbappend` file under `dynamic-layers/meta-python/...` to keep things grouped together?

agherzan: I would also like that. It feels like we just break that package when meta-python is not available.

kraj: Yes, I thought of that but this seems more directional when someone wanted to use this package that they would need meta-python, shoving all into dynamic-layers would make it more cryptic.

kraj: moved now

agherzan (approved):
Looks good!

agherzan:
@pbrkr Thumbs up?

pbrkr:
@agherzan Thumbs up from me. I'm currently testing a few changes which I've based on top of this branch.

pbrkr (approved):

agherzan:
Thanks, guys. Merged.

PR #773, opened 2020-12-01
cairo bbappend: remove egl with userland

twoerner:
When using the userland graphics (i.e. DISABLE_VC4GRAPHICS) cairo isn't able
to use the provided EGL:

	| checking for cairo's EGL functions feature...
	| checking whether cairo's EGL functions feature could be enabled... no (not required by any backend)
	| configure: error:  EGL functions feature could not be enabled

Signed-off-by: Trevor Woerner <twoerner@gmail.com>

PR #774, opened 2020-12-02
mesa-gl: Enable swrast dri backend for rpi/userland

kraj:
After mesa switched build system to meson, the logic to generate dri
related artifacts changed too, which means when no dri backend is
enabled then dri drivers and corresponding headers dont get generated
and hence we end up with missing pkgconfig files e.g. dri.pc which
usually will come from full mesa3d package, but in rpi when userland is
used we only build GL pieces of mesa.

This patch therefore enables swrast dri backend when using userland,
which gives us the needed header and .pc files to build packages like
Xorg server, since EGL driver will come from userland, the dri backend
would be unused and hence should not be effective at runtime.

Addresses issue report with https://github.com/agherzan/meta-raspberrypi/pull/773

Cc: Trevor Woerner <twoerner@gmail.com>

**- What I did**

**- How I did it**

### DIFF
--- a/recipes-graphics/mesa/mesa-gl_%.bbappend
+++ b/recipes-graphics/mesa/mesa-gl_%.bbappend
@@ -1,6 +1,8 @@
 PACKAGECONFIG_append_rpi = " gbm"
 PROVIDES_append_rpi = " virtual/libgbm"
 
+DRIDRIVERS_append_rpi = ",swrast"
+
 do_install_append_rpi() {
     rm -rf ${D}${includedir}/KHR/khrplatform.h
 }


twoerner:
There is no reason to turn off the gbm things, they are orthogonal to setting the DRI drivers to "swrast" with userland.

With a patch that only adds the "swrast" line, I now have glmark2-es2-dispmanx building and running (with various fixes to various recipes and the code). Interestingly, one does not need X11 to run glmark2-es2-dispmanx, dispmanx runs directly on the GPU with no window system/display server required.

agherzan (approved):

agherzan:
Thanks, guys. This looks good to me. Merged to master.

PR #776, opened 2020-12-04
raspidmx: add

twoerner:
Add a recipe to build a set of dispmanx examples.

Signed-off-by: Trevor Woerner <twoerner@gmail.com>

### DIFF
--- a/recipes-graphics/raspidmx/raspidmx/0001-gitignore-add-archives-from-lib-directory.patch
+++ b/recipes-graphics/raspidmx/raspidmx/0001-gitignore-add-archives-from-lib-directory.patch
@@ -1,0 +1,24 @@
+From 070b114fd54b6ad38b4eff04279fe89ccaaef61b Mon Sep 17 00:00:00 2001
+From: Trevor Woerner <twoerner@gmail.com>
+Date: Fri, 4 Dec 2020 01:39:30 -0500
+Subject: [PATCH] gitignore: add archives from lib directory
+
+The build creates two *.a files in the lib directory, add these to .gitignore.
+
+Upstream-status: submitted [https://github.com/AndrewFromMelbourne/raspidmx/pull/29]
+Signed-off-by: Trevor Woerner <twoerner@gmail.com>
+---
+ .gitignore | 1 +
+ 1 file changed, 1 insertion(+)
+
+diff --git a/.gitignore b/.gitignore
+index 633f325..4a1ca0c 100644
+--- a/.gitignore
++++ b/.gitignore
+@@ -21,5 +21,6 @@ test_pattern/test_pattern
+ worms/worms
+ 
+ # other directories
++lib/*.a
+ 
+ *_/
--- a/recipes-graphics/raspidmx/raspidmx_git.bb
+++ b/recipes-graphics/raspidmx/raspidmx_git.bb
@@ -1,0 +1,30 @@
+SUMMARY = "Some examples using the DispmanX API on the Raspberry Pi"
+HOMEPAGE = "https://github.com/AndrewFromMelbourne/raspidmx"
+SECTION = "graphics"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=52962875ab02c36df6cde47b1f463024"
+
+COMPATIBLE_HOST_rpi = "${@bb.utils.contains('MACHINE_FEATURES', 'vc4graphics', 'null', '(.*)', d)}"
+
+SRC_URI = "git://github.com/AndrewFromMelbourne/raspidmx;protocol=https \
+           file://0001-gitignore-add-archives-from-lib-directory.patch \
+           file://0002-add-install-targets-to-Makefiles.patch \
+           file://0003-switch-to-pkg-config.patch \
+           file://0004-add-libvchostif-to-link.patch \
+           file://0005-change-library-linking-order.patch \
+           file://0006-game-Makefile-install-sample-png-files.patch \
+           file://0007-Makefile-reorganize.patch \
+           "
+
+PV = "0.0+git${SRCPV}"
+SRCREV = "e2ee6faa0d01a5ece06bcc74a47f37d7e6837310"
+
+S = "${WORKDIR}/git"
+
+inherit pkgconfig
+
+DEPENDS += "libpng userland"
+
+do_install () {
+	oe_runmake 'DESTDIR=${D}' 'TARGET=install'
+}
--- a/recipes-graphics/raspidmx/raspidmx/0002-add-install-targets-to-Makefiles.patch
+++ b/recipes-graphics/raspidmx/raspidmx/0002-add-install-targets-to-Makefiles.patch
@@ -1,0 +1,214 @@
+From 3b60fa1598385fb36fef1c47f59cb08d3a6fba35 Mon Sep 17 00:00:00 2001
+From: Trevor Woerner <twoerner@gmail.com>
+Date: Fri, 4 Dec 2020 01:54:37 -0500
+Subject: [PATCH] add "install" targets to Makefiles
+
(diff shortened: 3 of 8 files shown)


recipes-graphics/raspidmx/raspidmx_git.bb
```diff
@@ -0,0 +1,36 @@
+SUMMARY = "A set of DispmanX examples"
```
agherzan, line 1: Let's use the same description in github as it clearly states the fact that it is RPi specific: Some examples using the DispmanX API on the Raspberry Pi .

recipes-graphics/raspidmx/raspidmx_git.bb
```diff
@@ -0,0 +1,36 @@
...
+           file://0001-gitignore-add-archives-from-lib-directory.patch \
+           file://0002-add-install-targets-to-Makefiles.patch \
+           file://0003-switch-to-pkg-config.patch \
+           file://0004-add-libvchostif-to-link.patch \
+           file://0005-change-library-linking-order.patch \
+           file://0006-game-Makefile-install-sample-png-files.patch \
+           file://0007-Makefile-reorganize.patch \
+           "
+
+PV = "1.0+git${SRCPV}"
```
agherzan, line 19: Any reason to set this to 1.0? I don't see that version in upstream and it might be a bit confusing.

recipes-graphics/raspidmx/raspidmx_git.bb
twoerner: Sounds good.

recipes-graphics/raspidmx/raspidmx_git.bb
twoerner: Agreed. It is a leftover from "devtool add…".
What do you recommend? Removing it? Setting it to "git${SRCPV}"? Something else?

agherzan: That or "0.0+git${SRCPV}". I saw that is used in other components in poky.

twoerner: Sounds good. I've re-pushed these two changes.

recipes-graphics/raspidmx/raspidmx_git.bb
```diff
@@ -0,0 +1,36 @@
...
+
+S = "${WORKDIR}/git"
+
+DEPENDS = "pkgconfig-native libpng userland"
+
+do_configure () {
+	:
+}
+
+do_compile () {
```
agherzan, line 30: What was the reason to overwrite this task?

recipes-graphics/raspidmx/raspidmx_git.bb
```diff
@@ -0,0 +1,36 @@
...
+           "
+
+PV = "0.0+git${SRCPV}"
+SRCREV = "e2ee6faa0d01a5ece06bcc74a47f37d7e6837310"
+
+S = "${WORKDIR}/git"
+
+DEPENDS = "pkgconfig-native libpng userland"
+
+do_configure () {
```
agherzan, line 26: `base_do_configure` also handles cleaning the package. I'm actually wondering if using `TARGET` os the right way for this.

recipes-graphics/raspidmx/raspidmx_git.bb
twoerner: it's what "devtool add…" gave me

agherzan: Uh yeah! You mentioned. Why would it do that? Strange.

recipes-graphics/raspidmx/raspidmx_git.bb
twoerner: The base Makefile that comes with the sources is pretty minimal, and doesn't clearly show the dependency on building "lib" first. When I re-wrote the Makefile I thought it would be fun to parallel-ize it making sure that no matter how it was built, "lib" was always first. One side-effect of the way that I parallel-ized it was that it needs this "TARGET=install" notation for "make install".

I could try other ways of re-writing the Makefile if you have preferences one way or another.

twoerner:
I've pushed my latest updates in response to your review. I appreciate your thoughts.

recipes-graphics/raspidmx/raspidmx_git.bb
agherzan: I"m fine with this. It would definitely be nicer to not need a TARGET - it's just not standard in any way.

agherzan (dismissed):

recipes-graphics/raspidmx/raspidmx_git.bb
agherzan: But let's leave that for further refinements as long as this works as expected.

kraj (commented):
please rebase instead of merging to pull branches

recipes-graphics/raspidmx/raspidmx/0001-gitignore-add-archives-from-lib-directory.patch
```diff
@@ -0,0 +1,24 @@
...
+
+diff --git a/.gitignore b/.gitignore
+index 633f325..4a1ca0c 100644
+--- a/.gitignore
++++ b/.gitignore
+@@ -21,5 +21,6 @@ test_pattern/test_pattern
+ worms/worms
+ 
+ # other directories
++lib/*a
```
kraj, line 22: this should be `*.a` isnt it ?

recipes-graphics/raspidmx/raspidmx_git.bb
```diff
@@ -0,0 +1,28 @@
...
+           file://0006-game-Makefile-install-sample-png-files.patch \
+           file://0007-Makefile-reorganize.patch \
+           "
+
+PV = "0.0+git${SRCPV}"
+SRCREV = "e2ee6faa0d01a5ece06bcc74a47f37d7e6837310"
+
+S = "${WORKDIR}/git"
+
+DEPENDS = "pkgconfig-native libpng userland"
```
kraj, line 24: this should be `inherit pkgconfig` and `pkgconfig-native` removed.

dev-0x7C6: I'm just curious, `DEPENDS =` should be changed to `DEPENDS +=` after `inherit pkgconfig`. Am I correct ?

kraj: usually classed append/prepend their dependencies so += does not matter but perhaps its safer yes.

recipes-graphics/raspidmx/raspidmx/0001-gitignore-add-archives-from-lib-directory.patch
twoerner: The only things that are in this directory are the Makefile, and the *.a files. So *a or *.a would be the same. But I can change it since *a might look suspicious.

recipes-graphics/raspidmx/raspidmx_git.bb
twoerner: I'll have to have a talk with the person who wrote devtool/recipetool…

recipes-graphics/raspidmx/raspidmx/0001-gitignore-add-archives-from-lib-directory.patch
kraj: yes lets use `*.a`

kraj (approved):
lgtm

twoerner:
Thank you for your comments and review :-)

agherzan (approved):
Thanks. Merged.

twoerner:
a small blog post about creating a small image with DISABLE_VC4GRAPHICS and a bunch of example apps:
https://twoerner.blogspot.com/2020/12/userland-graphics-with-openembeddedyocto.html

PR #780, opened 2020-12-14
raspidmx: Disable for all non-rpi hosts

kraj:
Currently its only disabled when vc4graphics is in use but this recipe
actually needs userland and therefore can not be used for non-rpi
machines as well.


agherzan (approved):
Looks good. Merged.

Issue #765, opened 2020-11-20
unable to boot raspberry pi 3b using 64 bit configuration


AnanduKE:
Hi,
I am unable to boot raspberry pi 3b using 64 bit configuration. Boot just hangs after
 Starting syslogd/klogd: done
i am using dunfell branch.
Enabled uart and currently using uart for getting console

agherzan:
Can you paste UART boot logs and also build configuration?

abhi-84:
I am facing the similar issue but I am not using UART. I ave tried with both dunfell and gatesgarth but same issue. With Warrior branch it is working fine. I wanted to access pi using ssh but ethernet port is not getting activated.
When connected with HDMI, I am getting following message:
[https://i.stack.imgur.com/dpaET.jpg](url)
Please help

Abhishek

agherzan:
That is not the same issue. Your boot fails while loading the kernel image in uboot while this issue reports hanging in userspace. As there is no update on the initial issue, I'll close this.

PR #783, opened 2020-12-17
Add overlay to allow 64 bit usage with CM3

shawaj:
Add overlay to allow 64 bit usage with CM3

Change-type: patch
Signed-off-by: Aaron Shaw <aaron@balena.io>

**- What I did**

Add CM3 overlay so you can use CM3 in 64 bit mode

**- How I did it**

Add overlay to conf file

### DIFF
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -53,8 +53,10 @@ RPI_KERNEL_DEVICETREE ?= " \
     bcm2710-rpi-3-b.dtb \
     bcm2710-rpi-3-b-plus.dtb \
     bcm2711-rpi-4-b.dtb \
+    bcm2711-rpi-400.dtb \
     bcm2708-rpi-cm.dtb \
     bcm2710-rpi-cm3.dtb \
+    bcm2711-rpi-cm4.dtb \
     "
 
 KERNEL_DEVICETREE ??= " \
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -46,10 +46,13 @@ RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
     "
 
 RPI_KERNEL_DEVICETREE ?= " \
+    bcm2708-rpi-zero.dtb \
     bcm2708-rpi-zero-w.dtb \
     bcm2708-rpi-b.dtb \
+    bcm2708-rpi-b-rev1.dtb \
     bcm2708-rpi-b-plus.dtb \
     bcm2709-rpi-2-b.dtb \
+    bcm2710-rpi-2-b.dtb \
     bcm2710-rpi-3-b.dtb \
     bcm2710-rpi-3-b-plus.dtb \
     bcm2711-rpi-4-b.dtb \
--- a/conf/machine/raspberrypi3-64.conf
+++ b/conf/machine/raspberrypi3-64.conf
@@ -18,6 +18,7 @@ RPI_KERNEL_DEVICETREE = " \
     broadcom/bcm2710-rpi-3-b.dtb \
     broadcom/bcm2710-rpi-3-b-plus.dtb \
     broadcom/bcm2837-rpi-3-b.dtb \
+    broadcom/bcm2710-rpi-cm3.dtb \
     "
 
 SERIAL_CONSOLES ?= "115200;ttyS0"
--- a/conf/machine/raspberrypi4-64.conf
+++ b/conf/machine/raspberrypi4-64.conf
@@ -15,6 +15,8 @@ include conf/machine/include/rpi-base.inc
 
 RPI_KERNEL_DEVICETREE = " \
     broadcom/bcm2711-rpi-4-b.dtb \
+    broadcom/bcm2711-rpi-400.dtb \
+    broadcom/bcm2711-rpi-cm4.dtb \
 "
 
 SDIMG_KERNELIMAGE ?= "kernel8.img"


shawaj:
Just realised that this closes https://github.com/agherzan/meta-raspberrypi/issues/765

Will amend commit message

kraj (commented):
While I am happy to have single machine serve both CM normal Pi4, I wonder if there are more things required for CM3 that would warrant a new machine conf

shawaj:
@kraj as far as I'm aware there's nothing else required. Literally just adding the dtb into boot directory allows the 64bit pi3 image to work on cm3

kraj:
> @kraj as far as I'm aware there's nothing else required. Literally just adding the dtb into boot directory allows the 64bit pi3 image to work on cm3

there is `raspberrypi-cm3.conf` for cm3 do we still need it then

shawaj:
@kraj that one is 32 bit so doesn't relate to this.

Or did you just mean that, we could also make amendments to merge cm3 32 bit one into the pi3 32 bit one?

Hope this makes sense.

kraj:
> @kraj that one is 32 bit so doesn't relate to this.
> 
> Or did you just mean that, we could also make amendments to merge cm3 32 bit one into the pi3 32 bit one?
> 
> Hope this makes sense.

right i was wondering that

shawaj:
@kraj yeah I guess you don't really need anything other than that dt overlay blob so you could effectively support cm1 just using raspberrypi.conf and cm3 using raspberrypi3.conf and cm3-64 using raspberrypi3-64.conf as far as I can tell

kraj:
> @kraj yeah I guess you don't really need anything other than that dt overlay blob so you could effectively support cm1 just using raspberrypi.conf and cm3 using raspberrypi3.conf and cm3-64 using raspberrypi3-64.conf as far as I can tell

would it be possible for you to try it out on 32bit version as well and perhaps retire cm3 image

agherzan:
I like that. I can't rememeber why we split them but I'm sure there were incompatibilities. Looking now where we are, I don't think that's the case anymore. Let's drop the cm ones. Actually, can we make them aliases so we don't break people's projects? And also, include a builds warning.

shawaj:
@agherzan yeah I think that's a good idea.

For the alias thing, do you have an example of what that looks like? And the builds warning too?

Happy to pr but my yocto knowledge is limited.

shawaj:
@kraj I managed to dig out my cm1 and dev kit so can test on all of them 😎

shawaj:
@kraj @agherzan just an update...I have tested the following:

- CM1 - works fine using any base image (dtb is included by default [here](https://github.com/agherzan/meta-raspberrypi/blob/8c0b7089fce5201ee45a04a6f614747fbefb3528/conf/machine/include/rpi-base.inc#L56))
- CM3 - flashed Pi 3 32 bit image, and it works straight out of the box ([as the dtb is already here](https://github.com/agherzan/meta-raspberrypi/blob/8c0b7089fce5201ee45a04a6f614747fbefb3528/conf/machine/include/rpi-base.inc#L57)) 
- CM3 64 bit - as discussed above and in this PR, it works perfectly by just bringing in the CM3 dtb to the raspberrypi3-64 image
- CM4 - we have no BSP for this yet as far as I am aware. But I guess we can do a similar thing with the Pi 4? The CM4 and Pi400 should all work using the Pi4 base image. We just need to pull in the new wifi drivers I believe as they both use the BCM43456 not the BCM43455 which is on the Pi 4

Based on this - what do you want to do? Merge PR and then remove CM3/CM1 specific BSPs and then amend the Pi4 one to support Pi400 and CM4?

kraj:
> @kraj @agherzan just an update...I have tested the following:
> 
> * CM1 - works fine using any base image (dtb is included by default [here](https://github.com/agherzan/meta-raspberrypi/blob/8c0b7089fce5201ee45a04a6f614747fbefb3528/conf/machine/include/rpi-base.inc#L56))
> * CM3 - flashed Pi 3 32 bit image, and it works straight out of the box ([as the dtb is already here](https://github.com/agherzan/meta-raspberrypi/blob/8c0b7089fce5201ee45a04a6f614747fbefb3528/conf/machine/include/rpi-base.inc#L57))
> * CM3 64 bit - as discussed above and in this PR, it works perfectly by just bringing in the CM3 dtb to the raspberrypi3-64 image
> * CM4 - we have no BSP for this yet as far as I am aware. But I guess we can do a similar thing with the Pi 4? The CM4 and Pi400 should all work using the Pi4 base image. We just need to pull in the new wifi drivers I believe as they both use the BCM43456 not the BCM43455 which is on the Pi 4
> 
> Based on this - what do you want to do? Merge PR and then remove CM3/CM1 specific BSPs and then amend the Pi4 one to support Pi400 and CM4?

Thanks for doing this work it’s really important and I think merging them is my suggestion so let’s have less machine variants as we can

agherzan:
@kraj Can we at least add a machine alias? I wouldn't like to break people's builds. I haven't done this before tbh but I reckon it should be possible to do.

kraj:
> @kraj Can we at least add a machine alias? I wouldn't like to break people's builds. I haven't done this before tbh but I reckon it should be possible to do.

yeah symlinks might be ok I think.

agherzan:
Sounds good. The only thing that won't be possible with symlinks is having a deprecation warning.

pbrkr:
@agherzan We could easily add deprecation warnings in another way. We could use http://git.yoctoproject.org/cgit/cgit.cgi/meta-virtualization/tree/classes/sanity-meta-virt.bbclass as an example and add a SanityCheck handler that issues a warning if MACHINE is set to one of the deprecated names.

agherzan:
@pbrkr Sure but the main idea was to remove the machines. I was looking for a way to have that warning while the machines reuse another machine. Needs some testing.

ryanteck:
Is there an ETA on when this will be merged? 

It's a feature I require.

Thanks

agherzan:
> Is there an ETA on when this will be merged?
> 
> It's a feature I require.

There are still items to handle (including, but not limited to, cleaning up the PR of merge commits, rebase etc.)

shawaj:
@agherzan @kraj I can clean up the PR and rebase no problem.

What do we want to do with the actual configs though? Merge the cm ones with the non cm ones and then have board aliases for the removed ones with deprecation warnings?

Do you have an example of what this looks like in yocto at all?

agherzan:
@shawaj Have you checked that all the kernel versions we currently have in the later support all these dts additions?

shawaj:
> @shawaj Have you checked that all the kernel versions we currently have in the later support all these dts additions?

@agherzan what exactly do you mean by this?

agherzan:
@shawaj `meta-raspberrypi` supports two kernel versions as of now (https://github.com/agherzan/meta-raspberrypi/tree/master/recipes-kernel/linux)/ Have you checked that both kernel versions provide the newly added dts?

shawaj:
@agherzan yes they are in both 5.4.y and 5.10.y

https://github.com/raspberrypi/linux/tree/rpi-5.4.y/arch/arm/boot/dts

https://github.com/raspberrypi/linux/tree/rpi-5.10.y/arch/arm/boot/dts

agherzan:
@agherzan-ci ok to test

shawaj:
@agherzan want me to merge the base branch changes or can you do that automatically?

agherzan (approved):

agherzan:
@shawaj all good. Merged.

shawaj:
Fantastic, thanks @agherzan 👍

PR #787, opened 2020-12-24
machine: use weak default for kernel and KERNEL_DEVICETREE settings

dozylynx:
**Updated PR description:**

To allow other layers to easily provide a default kernel preference to use with the Raspberry Pi MACHINEs, switch to use the `??=` assignment operator for setting the default kernel and the device tree value in this layer.

This change is motivated by enabling Xen support in meta-virtualization to use the raspberrypi4-64 MACHINE definition from this layer with the Yocto Linux kernels.

**The previous PR description with the original commits that were proposed, and have since been replaced, was:**
_machine: add vendorkernel to MACHINE_FEATURES for default kernel pref_

Xen support in the meta-virtualization layer uses the raspberrypi4-64
MACHINE definition from this layer with the Yocto Linux kernels rather
than the board-vendor-provided linux-raspberrypi kernel.

To improve support for selecting the kernel, introduce vendorkernel into
the available MACHINE_FEATURES: this sets the default kernel preference
and the inclusion of the Raspberry Pi kernel device tree overlays to
match the kernel. Add a DISABLE_VENDORKERNEL variable for ease of
toggling the default in meta-virtualization.

Since Xen support on the Raspberry Pi 4 board requires at least Linux
5.9, with 5.10 preferred, the non-vendor kernel default preference is
currently set to linux-yocto-dev.

### DIFF
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -57,7 +57,7 @@ RPI_KERNEL_DEVICETREE ?= " \
     bcm2710-rpi-cm3.dtb \
     "
 
-KERNEL_DEVICETREE ?= " \
+KERNEL_DEVICETREE ??= " \
     ${RPI_KERNEL_DEVICETREE} \
     ${RPI_KERNEL_DEVICETREE_OVERLAYS} \
     "
--- a/conf/machine/include/rpi-default-providers.inc
+++ b/conf/machine/include/rpi-default-providers.inc
@@ -1,6 +1,6 @@
 # RaspberryPi BSP default providers
 
-PREFERRED_PROVIDER_virtual/kernel ?= "linux-raspberrypi"
+PREFERRED_PROVIDER_virtual/kernel ??= "linux-raspberrypi"
 PREFERRED_PROVIDER_virtual/xserver ?= "xserver-xorg"
 PREFERRED_PROVIDER_virtual/egl ?= "${@bb.utils.contains("MACHINE_FEATURES", "vc4graphics", "mesa", "userland", d)}"
 PREFERRED_PROVIDER_virtual/libgles2 ?= "${@bb.utils.contains("MACHINE_FEATURES", "vc4graphics", "mesa", "userland", d)}"


agherzan:
@agherzan-ci ok to test

kraj (changes requested):
this change could jsut be done via your own layer to select correct kernel. why do we need this extra logic to complicate things ? I would rather prefer a new kernel like linux-mainline_5.10 which has full support for rpi now rather than these kind of tweaks.

agherzan:
I thought we hardcode the kernel provider. Checking it now, and seeing that this is not the case, I share @kraj 's opinion. Why can't this be done through an external layer/configuration? Is it because of the `KERNEL_DEVICETREE` entries?

dozylynx:
For context: I'm seeking to use the Yocto meta-raspberrypi layer as the provider of the MACHINE definition for the Raspberry Pi 4 board, and to enable it to be used with the Yocto meta-virtualization layer for running the Xen hypervisor, adding no further layers beyond their existing dependencies (from meta-openembedded).

The challenge with using an alternative kernel with the current meta-raspberrypi layer is indeed the `KERNEL_DEVICETREE` in the MACHINE definition, because it's encoding an assumption about the files produced by the kernel recipe, which then causes image recipes to fail to build when the kernel that is in use doesn't generate those overlay files. Since it's the MACHINE defintion that's introducing this behaviour, the MACHINE_FEATURES is a mechanism that can be used to adjust it when it needs to be done.

The linux-raspberrypi kernel has been tried with Xen and found to be unsuitable as a platform because of the regular breakage caused by the removal of source code revisions in the upstream Raspberry Pi project's kernel sources when they are rebased. The Yocto linux kernel that is provided by linux-yocto-dev recipes builds the mainline linux kernel, works on the hardware and is a suitable platform for collaboration, ultimately towards using the standard linux-yocto kernel.

pbrkr:
I'm not a fan of depending on `linux-yocto-dev` here, that recipe will follow pre-release kernels from upstream. I recommend instead just using `linux-yocto` from oe-core or `linux-stable` from the meta-linux-mainline layer if you want to use an upstream kernel. I already have an example of how to enable `linux-stable` for the `raspberrypi3` MACHINE in meta-linux-mainline, it should be trivial to adapt that to other Raspberry Pi variants.

pbrkr (changes requested):
I'm not a fan of depending on linux-yocto-dev here, that recipe will follow pre-release kernels from upstream. I recommend instead just using linux-yocto from oe-core or linux-stable from the meta-linux-mainline layer if you want to use an upstream kernel. I already have an example of how to enable linux-stable for the raspberrypi3 MACHINE in meta-linux-mainline, it should be trivial to adapt that to other Raspberry Pi variants.

dozylynx:
I share the general preference for `linux-yocto` over `linux-yocto-dev`, but `-dev` was selected for the time being since Linux 5.9 is the minimum required version for Xen compatibility on the Raspberry Pi 4, and current `linux-yocto` in oe-core is only at 5.8. If `linux-yocto` was ready, I would propose using it. With respect to other kernels, as I indicated above I am seeking to avoid introducing any further layer dependencies.

It looks like using an addition to `MACHINE_FEATURES` to enable clearing the population of the `KERNEL_DEVICETREE` entries is less contentious (and the necessary part of this change) than the selection of the specific alternative kernel in this layer, so I can respin this with just that change. The alternative kernel selection in meta-raspberrypi could be revisited when `linux-yocto` is updated with a newer kernel, but for now can be handled in meta-virtualization.

agherzan:
Let me understand something. So, this PR can be done by having a build configuration (distro, etc) which:

- switches the kernel provider to whatever is wanted (dev or not dev) 
- Nukes KERNEL_DEVICETREE

Wouldn't that work?

dozylynx:
> Let me understand something. So, this PR can be done by having a build configuration (distro, etc) which:
> 
> * switches the kernel provider to whatever is wanted (dev or not dev)
> * Nukes KERNEL_DEVICETREE
> 
> Wouldn't that work?

That is correct. I'm working through what the best way to do this is - thanks for your patience.

The `KERNEL_DEVICETREE` setting and kernel provider preference are set here in the `MACHINE` definition in this layer, and Xen guest support is enabled via an entry in `DISTRO_FEATURES` in meta-virtualization, where there are some other settings added that are specific to the Raspberry Pi 4 board - eg. adding the hypervisor binary to the SD card image and increasing partition size for that.

It could be helpful to be able to use the weak assignment operator `?=` to set the Xen-enabled kernel preference in meta-virtualization, to ensure that a user-supplied kernel preference (eg. in local.conf or some other layer) is able to easily take precedence over it; however, this layer already uses that operator so it doesn't quite have the intended effect when used again in meta-virtualization.

Would it be an acceptable proposal for this layer to adopt the "weak default value" operator `??=` instead to set the kernel preference and populate `KERNEL_DEVICETREE`, and allow meta-virtualization to use `?=` ? Do you think there would be any complications with that approach?

dozylynx:
I've pushed a revised commit to the branch for this PR: far simpler and hopefully acceptable

kraj (approved):
LGTM

agherzan:
@dozylynx That was exactly what I was aiming at. Thanks for this. Merged.

agherzan (approved):

PR #792, opened 2021-01-14
Enable use of 'RPI_EXTRA_CONFIG += ...' in local.conf

m5p3nc3r:
Enable the use of RPI_EXTRA_CONFIG when targeting 64bit images on PI4.

Before this patch, use of RPI_EXTRA_CONFIG += in a local.conf would overwrite
the needed 'arm_64bit=1' defined in the core recipes.  The net result would be
a system that did not boot.

This patch fixes the assignment of 'arm_64bit=1' even if RPI_EXTRA_CONFIG
has been set elsewhere.

Signed-off-by: Your Name matthew@thespencers.me.uk

### DIFF
--- a/conf/machine/raspberrypi4-64.conf
+++ b/conf/machine/raspberrypi4-64.conf
@@ -31,6 +31,6 @@ KERNEL_IMAGETYPE_UBOOT ?= "Image"
 KERNEL_IMAGETYPE_DIRECT ?= "Image"
 KERNEL_BOOTCMD ?= "booti"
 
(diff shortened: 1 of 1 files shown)


agherzan (changes requested):
The change looks fine to me. We would need to tweak the commit log in accordance with https://meta-raspberrypi.readthedocs.io/en/latest/contributing.html
See git history for examples.

m5p3nc3r:
Sorry, not sure what happened with the comment.  Updated, hopefully this is fine now?

agherzan:
@agherzan-ci ok to test

agherzan (approved):

PR #798, opened 2021-01-21
Kernel updates and minor config tweaks

pbrkr:
* Update the kernel & yocto-kernel-cache to the latest upstream commits.
* Use unified u-boot config for 64-bit targets.
* We no longer need to force `arm_64bit=1` to boot Raspberry Pi 4 in 64-bit mode.
* Tidy up kas builds a little: Place cloned layers under a `layers` subdirectory and ignore this from git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ build*
 *.rej
 *.log
 docs/_build
+/layers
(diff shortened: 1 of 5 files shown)


kraj (approved):
lgtm

agherzan (approved):
Looks good. Merged.

PR #803, opened 2021-01-30
rpi-config: comment updated

SCVready:
Hi,

I opened a thread in raspberrypi forum: https://www.raspberrypi.org/forums/viewtopic.php?f=66&t=298129 regarding the issue tackled in https://github.com/agherzan/meta-raspberrypi/commit/74deec51ded5b22b31b69b30537a6e9ed0997aea, and I got answered that there could be a "4K limit on reading config.txt in certain circumstances", and there was a similar case in which another variable was affected: https://github.com/raspberrypi/firmware/issues/1012

They also suggested to do not use the commented version file because it can very slightly increase the booting time as the firmware has more to parse.

Regards

### DIFF
--- a/recipes-bsp/bootfiles/rpi-config_git.bb
+++ b/recipes-bsp/bootfiles/rpi-config_git.bb
@@ -118,10 +118,15 @@ do_deploy() {
 
     # Video camera support
     if [ "${VIDEO_CAMERA}" = "1" ]; then
-        # TODO: It has been observed that Raspberry Pi 4B 4GB may fail to enable the camera if "start_x=1" is at the end
-        #       of the file. The underlying cause is unknown, but it can be related with a file size limitation affecting
-        #       this variable. Therefore, "start_x=1" has been set to replace the original occurrence in config.txt,
-        #       which is at the middle of the file.
+        #   It has been observed that Raspberry Pi 4B 4GB may fail to enable the
+        # camera if "start_x=1" is at the end of the file. Therefore,
+        # "start_x=1" has been set to replace the original occurrence in
+        # config.txt, which is at the middle of the file.
+        #   The exact underlying cause is unknown. There are similar issues
+        # reported in the raspberrypi/firware repo and the conclusion reached
+        # was that there could be a file size limitation affecting certain
+        # variables. It was commented that this limitation could be 4k but
+        # not proved.
         sed -i '/#start_x=/ c\start_x=1' $CONFIG
     fi
 


agherzan (changes requested):
I haven't read the entire thread, but I still see people talking about the size is still an issue. Can you please explain more what is your current understanding of the issue?

SCVready:
I think that this limitation almost certainly exist. There were at least another case where this problem was encountered, and the people tackling it also deduced the same thing. But, no one on those threads could explain why; maybe because it is hardware related? I cannot help more on this at the moment, I only have a little knowledge on the booting process of the rpi, and this issue is far beyond it

agherzan:
OK. So if that is the case, why do we remove the comment? @SCVready

SCVready:
Hi, I put the todo comment to check that the behave I saw was related with the size, and I think I confirmed it. But if you want clarify the issue beneath it, yeah this is not enough. I can close this pull request

agherzan:
@SCVready So what about rewording the comment based on the upstream feedback?

recipes-bsp/bootfiles/rpi-config_git.bb
```diff
@@ -118,10 +118,6 @@ do_deploy() {
 
     # Video camera support
     if [ "${VIDEO_CAMERA}" = "1" ]; then
-        # TODO: It has been observed that Raspberry Pi 4B 4GB may fail to enable the camera if "start_x=1" is at the end
-        #       of the file. The underlying cause is unknown, but it can be related with a file size limitation affecting
-        #       this variable. Therefore, "start_x=1" has been set to replace the original occurrence in config.txt,
```
kraj, line 123: these lines are more than 80 columns and this could be a problem in config parser to handle that see 

https://github.com/raspberrypi/firmware/issues/1012#issuecomment-401030535

Please reformat these comments so that they fit into < 80 chars on line in final config ( e.g. line 122 is 112 characters in length) 

see if that fixes the issue you have.

kraj (changes requested):

recipes-bsp/bootfiles/rpi-config_git.bb
agherzan: @SCVready any update on this?

SCVready: Oh sorry. Yes, I can update the comment and fit it in 80 chars columns. I'll update this pull request this weekend.

kraj (approved):
LGTM, you did not update if it fixed your problem or not. Does this fix the original issue you reported ?

SCVready:
Hi, the original issue was already fixed in my other commit. In this pull request I'm only modifying the comment I added with new info I found about the issue.

I also changed it to fit in 80 chars as you said, but won't make any difference as this comment is in the yocto recipe and it's not being transferred to the actual config.txt.

kraj:
> Hi, the original issue was already fixed in my other commit. In this pull request I'm only modifying the comment I added with new info I found about the issue.
> 
> I also changed it to fit in 80 chars as you said, but won't make any difference as this comment is in the yocto recipe and it's not being transferred to the actual config.txt.

so I want you to try to revert your other patch and see if this fixes the original problem by limiting comments to 80 chars

SCVready:
I checked that the config.txt we are using is already strictly limited to 80 chars (https://github.com/Evilpaul/RPi-config/blob/master/config.txt), and meta-raspberrypi is not adding any additional comment that exceed this limit either. So, this cannot be the cause

kraj:
> I checked that the config.txt we are using is already strictly limited to 80 chars (https://github.com/Evilpaul/RPi-config/blob/master/config.txt), and meta-raspberrypi is not adding any additional comment that exceed this limit either. So, this cannot be the cause

Is this the final config file that is on the rootfs ?
We generate the final config in do_deploy so please check the final config as seen in the image and report back

SCVready:
Hi, Checking the config file that is burned in the image I don't see any line that exceed this limit

kraj:
> Hi, Checking the config file that is burned in the image I don't see any line that exceed this limit

yes its no longer in generated `config.txt` here as well

agherzan:
I am a bit confused by this thread guys. Can any of you shed some light on it so we can bring it to a closure?

kraj:
> I am a bit confused by this thread guys. Can any of you shed some light on it so we can bring it to a closure?

I was suspecting that file has lines exceeding 80 chars which will trip the poor parser due to that comment in code but these comments are absent from final config.txt that’s installed on rootfs so that theory have failed 

another theory is that file is exceeding max size which is iirc 4K or something we have not tested the theory 

In summary the issue is still at large AFAICT

agherzan:
> > I am a bit confused by this thread guys. Can any of you shed some light on it so we can bring it to a closure?
> 
> I was suspecting that file has lines exceeding 80 chars which will trip the poor parser due to that comment in code but these comments are absent from final config.txt that’s installed on rootfs so that theory have failed
> 
> another theory is that file is exceeding max size which is iirc 4K or something we have not tested the theory
> 
> In summary the issue is still at large AFAICT

Right. That was my understanding but the confusing part is that we say that we remove the comment in this PR. It seems we only reword it a bit.

kraj:
> > > I am a bit confused by this thread guys. Can any of you shed some light on it so we can bring it to a closure?
> > 
> > 
> > I was suspecting that file has lines exceeding 80 chars which will trip the poor parser due to that comment in code but these comments are absent from final config.txt that’s installed on rootfs so that theory have failed
> > another theory is that file is exceeding max size which is iirc 4K or something we have not tested the theory
> > In summary the issue is still at large AFAICT
> 
> Right. That was my understanding but the confusing part is that we say that we remove the comment in this PR. It seems we only reword it a bit.

yeah rewording is good in itself. So this patch could be accepted as an improvement to general health but perhaps is not
fix for original problem.

agherzan (approved):

agherzan:
@agherzan-ci ok to test

kraj:
@agherzan can we merge this regardless

agherzan (approved):

agherzan:
@kraj I fail to see the improvement, to be honest, but I don't mind merging it. The only real difference I see is the text wrapping unless I'm missing something.

kraj:
> @kraj I fail to see the improvement, to be honest, but I don't mind merging it. The only real difference I see is the text wrapping unless I'm missing something.

me too. I was looking to close it out in whichever way

PR #808, opened 2021-02-05
Update kernel and firmware to latest 5.10.13

shr-project:


kraj (approved):
lgtm

shr-project:
Build tested locally for rpi3, rpi4, rpi4-64 and runtime tested with webOS on rpi4.

Seems to work at least as good as 5.4 did :). But more runtime testing would be useful I guess or we can merge it without changing  the default version now, but then we would need to verify that new firmware is still working fine with 5.4 kernel.

agherzan (approved):

PR #815, opened 2021-02-21
Add recipes for most of the Adafruit CircuitPython Python libraries

mostthingsweb:
This makes it easy to use various Adafruit products, such as the DC & Stepper Motor Bonnet.

PR #818, opened 2021-02-27
python3-adafruit-blinka: Make rpi-gpio dependency conditional

kraj:
https://github.com/agherzan/meta-raspberrypi/pull/815

introduced python3-adafruit-blinka and this recipe has added rdep on rpi-gpio,
however rpi-gpio is rpi specific package, and therefore should only be added
when building rpi based platforms, this makes the layer work in a
multi-bsp setup work

**- What I did**

**- How I did it**

### DIFF
--- a/recipes-devtools/python/python3-adafruit-blinka_6.2.2.bb
+++ b/recipes-devtools/python/python3-adafruit-blinka_6.2.2.bb
@@ -17,5 +17,6 @@ RDEPENDS_${PN} += " \
     python3-adafruit-platformdetect \
     python3-adafruit-pureio \
     python3-core \
-    rpi-gpio \
 "
+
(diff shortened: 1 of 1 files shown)


agherzan:
But is modules also rpi specific? Maybe the entire package should be only for rpi.

kraj:
> But is modules also rpi specific? Maybe the entire package should be only for rpi.

they are not pi specific

agherzan (approved):

agherzan:
@kraj of that is the case, maybe we should push them to meta-pyrhob instead.

agherzan:
I've merged this to unbreak builds.

kraj:
Thanks, it ubborked the 32bit builds but saw that 64bit builds were still breaking with different errors. Root caused it and sent another fix on top of this one. Hopefully this will put it to rest.

PR #819, opened 2021-02-27
python3-adafruit-blinka: Make it available only for 32bit rpi machines

kraj:
It has prebuilt binary libgpiod_pulsein for rpi machines and these binary is 32bit
therefore we can not include it for 64bit machines even if they are rpi
based unless they have multilib enabled.

This patch makes it visible only on rpi-32bit

Fixes QA errors like
ERROR: QA Issue: Architecture did not match (ARM, expected AArch64) in /usr/lib/python3.9/site-packages/adafruit_blinka/microcontroller/bcm283x/pulseio/.debug/libgpiod_pulsein [arch]

### DIFF
--- a/recipes-devtools/python/python3-adafruit-blinka_6.2.2.bb
+++ b/recipes-devtools/python/python3-adafruit-blinka_6.2.2.bb
@@ -12,6 +12,14 @@ inherit setuptools3
 
 DEPENDS += "python3-setuptools-scm-native"
 
+do_install_append() {
+# it ships ./bcm283x/pulseio/libgpiod_pulsein which is a prebuilt
+# 32bit binary therefore we should make this specific to 32bit rpi machines (based on bcm283x) only
+    if [ ${@bb.utils.contains('TUNE_FEATURES', 'callconvention-hard', '1', '0', d)} = "0" ]; then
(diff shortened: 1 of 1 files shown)


kraj:
@agherzan you were in right direction, we needed additional fix since some of the microcontroller implementations have prebuilt binaries for rpi 32bit. This patch here should fix it for good finally.

recipes-devtools/python/python3-adafruit-blinka_6.2.2.bb
```diff
@@ -12,6 +12,14 @@ inherit setuptools3
 
 DEPENDS += "python3-setuptools-scm-native"
 
+do_install_append() {
```
agherzan, line 15: How does this end up RPI specific? The conditional seems to only be on hard float. which indeed is always the case for armv8. Should this me a rpi append?

kraj: yeah its 32bit armhf specific since thats what the binary is for. the use of that binary is for SOCs which are in rpi boards 
I kept it with wider cut. Although this recipe has microcontroller specific implementations but that's a runtme thing

agherzan (approved):
Merged. Thanks @kraj

PR #820, opened 2021-03-03
Kernel 5.10.17

MikeLooijmans:
**- What I did**

Fixes issues with the ARM-based KMS driver on the RPi4

**- How I did it**

Upgrade kernel and firmware to the current state of affairs to fix regression in the I2C controller.
Add missing overlay files to the boot partition. In particular "overlay_map" needed to be in the overlays folder.

### DIFF
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -15,6 +15,7 @@ XSERVER = " \
     "
 
 RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
+    overlays/overlay_map.dtb \
     overlays/at86rf233.dtbo \
     overlays/disable-bt.dtbo \
     overlays/dwc2.dtbo \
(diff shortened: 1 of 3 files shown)


conf/machine/include/rpi-base.inc
```diff
@@ -15,6 +15,7 @@ XSERVER = " \
     "
 
 RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
+    overlays/overlay_map.dtb \
```
agherzan, line 18: The log doesn't explain what this actually is. Could you please expand. Also, the commit misses SoB.

agherzan (changes requested):

conf/machine/include/rpi-base.inc
MikeLooijmans: Added the SoB and a link to the detailed explanation of what overlay_map is for.

kraj: does this go into overlays/ folder ?

MikeLooijmans: Yes. Hence some extra code in the recipe to put this particular file in the overlays folder and not in the root like the rest of the dtb files.
I tested it on the rpi4 board and confirmed that the firmware will only use it when placed in the overlays folder.

agherzan: Thanks for clarifying. I'm good with this.

agherzan (approved):

agherzan:
@agherzan-ci ok to test

MikeLooijmans:
Rebased on current master....

agherzan (approved):

agherzan:
Thanks, @MikeLooijmans. Merged.

PR #825, opened 2021-03-10
rename gstreamer1.0-plugins-good_1.18.%.bbappend for gatesgarth

MikeLooijmans:
layer.conf lists "gatesgarth" as a compatible branch, but that does not
actually work because gatesgarth uses gstreamer 1.16 while hardknott will
use 1.18. To at least be able to pass the recipe-parsing phase, rename
the bbappend to have a wildcard that matches both revisions.


ricardosalveti:
@MikeLooijmans created https://github.com/agherzan/meta-raspberrypi/pull/829 for the backporting effort, mind having a look there as well?

agherzan:
I would rather have `gatesgarth` removed from `LAYERSERIES_COMPAT` for which the specific branch is to be used.

quaresmajose:
The packge config on this this bbappend only works with the master oe-core and with the PR https://github.com/agherzan/meta-raspberrypi/pull/832 gatesgarth will be removed from the  compatible branch

agherzan (changes requested):
It looks good but please tweak the git of the subject line. See git log in the respective directory for example or check out https://meta-raspberrypi.readthedocs.io/en/latest/contributing.html

agherzan:
I guess we can close this given that `LAYERSERIES_COMPAT` doesn't advertise `gatesgarth` anymore.

PR #854, opened 2021-05-09
rpi-base: make SPLASH overridable from outside

texierp:
Signed-off-by: Pierre-Jean Texier <texier.pj2@gmail.com>

### DIFF
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -102,7 +102,7 @@ SERIAL_CONSOLES_CHECK ??= "${SERIAL_CONSOLES}"
 BOOTFILES_DIR_NAME ?= "bootfiles"
 
 # Set Raspberrypi splash image
-SPLASH = "psplash-raspberrypi"
+SPLASH ?= "psplash-raspberrypi"
 
 def make_dtb_boot_files(d):
     # Generate IMAGE_BOOT_FILES entries for device tree files listed in


conf/machine/include/rpi-base.inc
```diff
@@ -102,7 +102,7 @@ SERIAL_CONSOLES_CHECK ??= "${SERIAL_CONSOLES}"
 BOOTFILES_DIR_NAME ?= "bootfiles"
 
 # Set Raspberrypi splash image
-SPLASH = "psplash-raspberrypi"
+SPLASH ?= "psplash-raspberrypi"
```
kraj, line 105: oe-core also has `meta/classes/image.bbclass` `SPLASH ?= "psplash"` have you tested which one gets preferred when external setting is not in play ?

texierp: Thanks @kraj for the review.

Yes, checked on my side. And when `SPLASH ?= "psplash-raspberrypi"` is like this, so, overridable and without an external setting (e.g `SPLASH = foo`), `psplash-raspberry` is preferred.

Otherwise, for safe version, I can introduce a new global variable `RPI_SPLASH_NAME` or something else  ... (default = `RPI_SPLASH_NAME ?= "psplash-raspberrypi" `) and update the `SPLASH` variable as follow:

`SPLASH = "${RPI_SPLASH_NAME}" `

Let me know.

kraj: no need for new variable. We want to use core variable as much as possible. If the order is

custom > meta-rpi > oe-core  then I am ok

texierp: Sure, you're right @kraj, so in this case, this PR is working as expected ;)

PR #856, opened 2021-05-13
99-com.rules: fix error invalid substitution type

sandy-lcq:
Fix below error:
/etc/udev/rules.d/99-com.rules:10 Invalid value "/bin/sh -c
'ALIASES=/proc/device-tree/aliases; if cmp -s $ALIASES/uart0
$ALIASES/serial0; then echo 0;elif cmp -s $ALIASES/uart0
$ALIASES/serial1; then echo 1; else exit 1; fi'" for PROGRAM (char 58:
invalid substitution type)

Signed-off-by: Changqing Li <changqing.li@windriver.com>


agherzan (commented):
Interesting. I'm wondering how come this was missed initially. Did something change in udev?

agherzan:
@agherzan-ci ok to test

kraj (approved):
lgtm

agherzan (approved):

PR #858, opened 2021-05-14
Update rpi distro firmwares

kraj:


agherzan (changes requested):
License seems to have changed.

agherzan (approved):

PR #861, opened 2021-05-24
Add WM8960 soundcard support

tewarid:
Add support for wm8960 based sound cards such as the WM8960 Hi-Fi Sound Card HAT for Raspberry Pi from Waveshare, and ReSpeaker 2 / 4 / 6 Mics Pi HAT from Seeed Studio.


recipes-kernel/linux/files/wm8960.cfg
```diff
@@ -0,0 +1,2 @@
+CONFIG_I2C_BCM2835=y
+CONFIG_SND_SOC_WM8960=m
```
kraj, line 2: should it be a module ?

tewarid: Could you point to a module example to clarify?

tewarid:
> Please rebase it on top of master and push again

Done.

recipes-bsp/bootfiles/rpi-config_git.bb
```diff
@@ -30,6 +30,9 @@ GPIO_IR_TX ?= "17"
 
 CAN_OSCILLATOR ?= "16000000"
 
+WM8960="${@bb.utils.contains("MACHINE_FEATURES", "wm8960", "1", "0", d)}"
```
agherzan, line 33: This deviates from the rest of the dtbo enablement in this recipe. None use machine features. And strictly, I'm not sure if a hat is a machine feature. I'd keep it for now as a separate variable and maybe unify it in the future somehow.

Also, we need to update the docs for this support (see existing docs).

recipes-bsp/bootfiles/rpi-config_git.bb
```diff
@@ -30,6 +30,9 @@ GPIO_IR_TX ?= "17"
...
+WM8960DTBO="wm8960-soundcard"
```
agherzan, line 34: This feels like an undeeded extra variable.

recipes-kernel/linux/files/wm8960.cfg
agherzan: @tewarid kraj is asking if this kconfig should really be a module and not built-in. If it is a module, we need to make sure that the kernel module ends up in the rootfs as well.

agherzan (changes requested):
And extra NIT: capitalise your git log subject line.

recipes-kernel/linux/files/wm8960.cfg
tewarid: @agherzan Thanks for the clarification :)

I'll make it built-in.

recipes-bsp/bootfiles/rpi-config_git.bb
tewarid: Is `WM8960DTBO` any different from `VC4DTBO`? I shall eliminate it.

recipes-bsp/bootfiles/rpi-config_git.bb
tewarid: I'll update the docs.

recipes-bsp/bootfiles/rpi-config_git.bb
agherzan: Yes, it is because that is using a weak assignment leaving the user the ability to overwrite. I don't see that as relevant in this case. Or is it?

tewarid: Makes sense now. There's no need for wm8960-soundcard dtbo to be overridden.

recipes-bsp/bootfiles/rpi-config_git.bb
tewarid: @agherzan Why is WM8960 any different than PITFT?

agherzan: I see. I've forgotten we do that for PITFT too. OK. I'm fine with the machine feature as well.

agherzan:
@agherzan-ci ok to test

agherzan (approved):

agherzan:
Thanks. Merged.

PR #863, opened 2021-05-25
layer.conf: Define LAYERDEPENDS

agherzan:


conf/layer.conf
```diff
@@ -10,6 +10,7 @@ BBFILE_PATTERN_raspberrypi := "^${LAYERDIR}/"
 BBFILE_PRIORITY_raspberrypi = "9"
 
 LAYERSERIES_COMPAT_raspberrypi = "hardknott honister"
+LAYERDEPENDS_raspberrypi = "core"
```
kraj, line 13: dont me require meta-python and meta-oe as well ?

agherzan: I'll give it a sping but I think we isolated those deps. Let's see...

agherzan: Parsing succeeds on `core-image-base` without anything but:

```
layer                 path                                      priority
==========================================================================
meta                  /home/andrei/work/yocto/poky/meta         5
meta-poky             /home/andrei/work/yocto/poky/meta-poky    5
meta-yocto-bsp        /home/andrei/work/yocto/poky/meta-yocto-bsp  5
meta-raspberrypi      /home/andrei/work/yocto/meta-raspberrypi  9
```

kraj: does it build too ?

agherzan: 2003/4240

What what do you suspect would be the problem? I would have expected parsing to surface a dangling bbappend or so.

agherzan: All good on master.

kraj (approved):
LGTM

PR #886, opened 2021-07-30
Convert to new override syntax

shr-project:


kraj (approved):
lgtm

agherzan:
@shr-project do you mind merging the two commits on syntax change?

shr-project:
> @shr-project do you mind merging the two commits on syntax change?

you mean squashing them? Please no as now the first script is created only by the script and the 2nd small is manually created and keeping them separate would help reviewing changes if the first one ever needs to be re-created with updated script.

agherzan:
That makes sense @shr-project. Let's merge this. CI will keep failing until meta-oe catches up.

conf/layer.conf
```diff
@@ -9,7 +9,7 @@ BBFILE_COLLECTIONS += "raspberrypi"
 BBFILE_PATTERN_raspberrypi := "^${LAYERDIR}/"
 BBFILE_PRIORITY_raspberrypi = "9"
 
-LAYERSERIES_COMPAT_raspberrypi = "hardknott honister"
```
agherzan, line 12: So in this case it is not an override?

conf/layer.conf
```diff
@@ -36,5 +36,5 @@ BBFILES_DYNAMIC += " \
     multimedia-layer:${LAYERDIR}/dynamic-layers/multimedia-layer/*/*/*.bbappend \
 "
 
-DEFAULT_TEST_SUITES_remove_rpi = "parselogs"
-DEFAULT_TEST_SUITES_append_rpi = " parselogs_rpi"
+DEFAULT_TEST_SUITES:remove:rpi = "parselogs"
+DEFAULT_TEST_SUITES:append:rpi = " parselogs:rpi"
```
agherzan, line 40: How does this one work?

agherzan: I think the testcase is called "parselogs_rpi".

conf/layer.conf
agherzan: re: LAYERSERIES_COMPAT

agherzan (changes requested):
Added 2 comments.

conf/layer.conf
shr-project: will fix

conf/layer.conf
shr-project: no, here "raspberrypi" is the name of the layer, not OVERRIDE.

agherzan: So a collection is not an override? I always thought it is.

agherzan (approved):
Looks good now. I think we can merge as CI will still fail until meta-oe catches up.

PR #888, opened 2021-07-30
Recipe updates

kraj:


conf/layer.conf
```diff
@@ -36,5 +36,5 @@ BBFILES_DYNAMIC += " \
     multimedia-layer:${LAYERDIR}/dynamic-layers/multimedia-layer/*/*/*.bbappend \
 "
 
-DEFAULT_TEST_SUITES_remove_rpi = "parselogs"
-DEFAULT_TEST_SUITES_append_rpi = " parselogs_rpi"
+DEFAULT_TEST_SUITES:remove:rpi = "parselogs"
```
agherzan, line 39: Do you have a link to the documentation of this change? I wasn't aware of it

kraj: sorry, I think I will open another pull. These changes are not yet needed, since they are only in master-next of oe-core as of now.

kraj:
closing

conf/layer.conf
agherzan: I don't think that is the case. There are already in master as far as I can see. I'll fix em.

PR #901, opened 2021-08-13
xserver-xf86-config: Correctly append to FILES:${PN}

trini:
When updating FILES:${PN} based on an override you must use
FILES:${PN}:append:OVERRIDE as the syntax otherwise you will end up
replacing the contents of FILES:${PN} entirely.  Update to use this
syntax correctly and then only add "${sysconfdir}/X11/xorg.conf.d/*"

Reported-by: Quentin Schulz <quentin.schulz@theobroma-systems.com>
Signed-off-by: Tom Rini <trini@konsulko.com>

**- What I did**

**- How I did it**

### DIFF
--- a/recipes-graphics/xorg-xserver/xserver-xf86-config_%.bbappend
+++ b/recipes-graphics/xorg-xserver/xserver-xf86-config_%.bbappend
@@ -13,4 +13,4 @@ do_install:append:rpi () {
     fi
 }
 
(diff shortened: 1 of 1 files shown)


kraj (approved):
good find. I wonder how many places we have used _append and += together, they would need fixing too

trini:
I did audit for other `FILES` cases but not more generally.

agherzan:
@agherzan-ci ok to test

PR #903, opened 2021-08-19
Revert "linux-raspberry: Drop 64-bit specific do_compile_append()"

floion:
This reverts commit a7cc636d4ef0ed7ddabf5785463dbb5c79633b1e which broke
loading overlays. The behaviour is that, as per the comments in
6c4de0b5fe44b8e661f1391ee8540a7f04d75315, the device tree blobs
generated on 64 bits lack the symbols and thus no overlays can be
loaded (checked with running "vcdbg log msg"):

005284.623: dtb_file 'bcm2711-rpi-4-b.dtb'
005289.625: brfs: File read: /mfs/sd/bcm2711-rpi-4-b.dtb
005289.641: Loading 'bcm2711-rpi-4-b.dtb' to 0x100 size 0x8c60
005302.330: brfs: File read: 35936 bytes
005310.499: brfs: File read: /mfs/sd/overlays/overlay_map.dtb
005358.309: brfs: File read: 1559 bytes
005360.894: brfs: File read: /mfs/sd/config.txt
005361.275: brfs: File read: 143 bytes
005372.465: brfs: File read: /mfs/sd/overlays/vc4-fkms-v3d.dtbo
005375.749: dterror: no symbols found
005375.760: Failed to resolve overlay 'vc4-fkms-v3d'
005376.120: brfs: File read: 1446 bytes
005385.470: brfs: File read: /mfs/sd/overlays/disable-wifi.dtbo
005388.352: dterror: no symbols found
005388.361: Failed to resolve overlay 'disable-wifi'

**- What I did**

**- How I did it**

### DIFF
--- a/recipes-kernel/linux/linux-raspberrypi.inc
+++ b/recipes-kernel/linux/linux-raspberrypi.inc
@@ -35,3 +35,10 @@ KERNEL_MODULE_AUTOLOAD += "${@bb.utils.contains("MACHINE_FEATURES", "pitft28r", 
 # set by default in rpi-4.8.y and later branches so we need to provide it
 # manually. This value unused if KERNEL_IMAGETYPE is not uImage.
 KERNEL_EXTRA_ARGS += "LOADADDR=0x00008000"
+
+do_compile_append() {
+    if [ "${SITEINFO_BITS}" = "64" ]; then
+        cc_extra=$(get_cc_option)
+        oe_runmake dtbs CC="${KERNEL_CC} $cc_extra " LD="${KERNEL_LD}" ${KERNEL_EXTRA_ARGS}
+    fi
+}


recipes-kernel/linux/linux-raspberrypi.inc
```diff
@@ -35,3 +35,10 @@ KERNEL_MODULE_AUTOLOAD += "${@bb.utils.contains("MACHINE_FEATURES", "pitft28r",
 # set by default in rpi-4.8.y and later branches so we need to provide it
 # manually. This value unused if KERNEL_IMAGETYPE is not uImage.
 KERNEL_EXTRA_ARGS += "LOADADDR=0x00008000"
+
+do_compile_append() {
+    if [ "${SITEINFO_BITS}" = "64" ]; then
+        cc_extra=$(get_cc_option)
+        oe_runmake dtbs CC="${KERNEL_CC} $cc_extra " LD="${KERNEL_LD}" ${KERNEL_EXTRA_ARGS}
+    fi
+}
```
kraj, line 44: I am not sure how this is fixing the problem but it’s looking to paper over the real issue since the commit to remove it tested the checksums of generated DTB files and they were identical so something else has changed 

please do a build with verbose on and post the command lines for these dtbs I think there is strip happening at sometime during dtb build and that’s where the problem is

kraj (changes requested):

recipes-kernel/linux/linux-raspberrypi.inc
floion: I don't know how the md5sums were identical for the committer of the changes since in my tests they are not. There is a huge size difference of the generated dtbs with or without this patch.
The dtbs are not stripped. The problem is that actually without this revert, this Makefile is not even considered https://github.com/raspberrypi/linux/blob/rpi-5.10.y/arch/arm64/boot/dts/broadcom/Makefile. So because it is not considered when building the dtbs, then https://github.com/raspberrypi/linux/blob/rpi-5.10.y/arch/arm64/boot/dts/broadcom/Makefile#L19 won't be taken into consideration. And that's why no symbols will be generated.
You can do a quick test by fetching one of your CI generated images, such as https://yocto-ci.resin.io/job/meta-raspberrypi-generic/lastSuccessfulBuild/artifact/deploy/core-image-base-raspberrypi3-64.wic.bz2 then add the `disable-wifi` and check that it won't disable the wifi. And then if you install the vc bins, you can check it complains about the symbols when running `vcdbg log msg`.

agherzan: OK. So looking a bit into this it looks like everything is hanging on symbols generation. I remember poky was also generating symbols based on the overlay detection function. Checking code, it is `devicetree_source_is_overlay`. @floion do you have the dtc full compile line (with flags) used as of now (without this revert)?

agherzan: I've just given this is a small test and:
1. the build compile the overlays with symbols
2. the symbols survive down to deploy

I've checked this on rpi4-64.

agherzan: I've done a similar test on rpi3-64 too. I haven't booted the images to triple check this but the overlays have symbols.

floion: @agherzan the problem is not with the overlays. The problem is on the rpi3 and rpi4 dtbs. So the device tree blobs, not with the additional overlays.

floion: And if you look at this thread for example, here specifically: https://lore.kernel.org/patchwork/patch/821645/#1016096 those guys state that the dtbs should actually be compiled like:
> "make DTC_FLAGS='-@' dtbs" should already work

And that's what this revert brings back actually, `make dtbs`

The commit https://github.com/agherzan/meta-raspberrypi/commit/a7cc636d4ef0ed7ddabf5785463dbb5c79633b1e says that the flags have been moved to kernel-devicetree.bbclass as per https://github.com/openembedded/openembedded-core/commit/74619de02774 but I think the issue is not the flags here but rather the make target that needs to be used.

agherzan: You're initial revert reads "This reverts commit a7cc636 which broke loading overlays. ". Could you clarify what broke exactly?

agherzan: re dtbs target. There is nothing special about that target. It is syntactic sugar for building all the dtb/dtbo for the enabled configuration. So, building on a per-file target or with `dtbs` should result in the same output (assuming the same arguments).

agherzan: @floion I've tried this at runtime too. With `disable-wifi` enabled, the WiFi interface doesn't exist:

```
root@raspberrypi4-64:~# tail  -n2 /mnt/config.txt
dtoverlay=disable-wifi
root@raspberrypi4-64:~# ifconfig wlan0
ifconfig: wlan0: error fetching interface information: Device not found
```

Without enabling the above overlay (commented out below), the interface is enabled:

```
root@raspberrypi4-64:~# tail  -n2 /mnt/config.txt
#dtoverlay=disable-wifi
root@raspberrypi4-64:~# ifconfig wlan0
root@raspberrypi4-64:~# ifconfig wlan0
wlan0     Link encap:Ethernet  HWaddr DC:A6:32:01:2D:25
          BROADCAST MULTICAST  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)
```

This was done using the RaspberryPi 4 (64) machine with the current HEAD of the `master` branch.

It looks like everything works as expected. Am I missing something?

agherzan (changes requested):
Not reproducible on RPi 4 (64). Waiting for feedback.

recipes-kernel/linux/linux-raspberrypi.inc
floion: I have reproduced it with https://yocto-ci.resin.io/job/meta-raspberrypi-generic/lastSuccessfulBuild/artifact/deploy/core-image-base-raspberrypi3-64.wic.bz2
Did you use a CI generated image or you built master? If you built master locally I'll give that a try also and check it out

agherzan: Indeed. I've used master poky/meta-rpi.

agherzan: @floion did you manage to reproduce it on master?

agherzan: @floion do you have any updates on it. I'd like to close this MR while we point the actual issue you are facing.

I've lately tested this on dunfell and it works as expected.

PR #932, opened 2021-11-03
dunfell: 99-com.rules: fix error invalid substitution type

landgraf:
fix below error:
/etc/udev/rules.d/99-com.rules:10 Invalid value "/bin/sh -c
'ALIASES=/proc/device-tree/aliases; if cmp -s $ALIASES/uart0
$ALIASES/serial0; then echo 0;elif cmp -s $ALIASES/uart0
$ALIASES/serial1; then echo 1; else exit 1; fi'" for PROGRAM (char 58:
invalid substitution type)


agherzan:
@agherzan-ci ok to test

agherzan (approved):

PR #941, opened 2021-11-13
rpi-config: Take into consideration ENABLE_UART value of 0

agherzan:
Also, validate if the value of it is not 0 or 1.

Fixes: https://github.com/agherzan/meta-raspberrypi/issues/567

### DIFF
--- a/recipes-bsp/bootfiles/rpi-config_git.bb
+++ b/recipes-bsp/bootfiles/rpi-config_git.bb
@@ -174,9 +174,11 @@ do_deploy() {
     fi
 
     # UART support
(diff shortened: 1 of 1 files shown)


kraj (approved):
lgtm

shr-project:
What about the empty value, which is the default? This is now causing:
```
rpi-config-git-r5 do_deploy: Invalid value for ENABLE_UART []. The value for ENABLE_UART can be 0 or 1.
```

agherzan:
@shr-project That is an issue. I'll fix it.

Issue #893, opened 2021-08-04
libwpe workaround for rpi to be removed from OE-Core


denix0:
Please migrate the corresponding workaround code for libwpe to meta-raspberrypi, if still applicable, as it is to be removed from OE-Core.

    # Workaround build issue with RPi userland EGL libraries.
    CFLAGS:append_rpi = " ${@bb.utils.contains('MACHINE_FEATURES', 'vc4graphics', '', '-D_GNU_SOURCE', d)}"

https://lists.openembedded.org/g/openembedded-core/topic/84653556

agherzan:
I think it is safe to go ahead and migrate. I'm curious why was it done there in the first place.

PR #942, opened 2021-11-14
libwpe: Migrate build workaround from oe-core

agherzan:
This was removed from oe-core[1] so we pull in the change here where it
should have been in the first place.

Fixes: https://github.com/agherzan/meta-raspberrypi/issues/893

[1] https://lists.openembedded.org/g/openembedded-core/topic/84653556


kraj (approved):
lgtm, thanks for fixing it

PR #961, opened 2021-11-29
weston-init: Add --continue-without-input option to weston

kraj:
launching with --continue-without-input helps starting weston
without keyboard/mouse

### DIFF
--- a/recipes-graphics/wayland/weston-init.bbappend
+++ b/recipes-graphics/wayland/weston-init.bbappend
@@ -1,0 +1,10 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
+do_install:append:rpi() {
+    if [ -e ${D}/${sysconfdir}/init.d/weston ]; then
+        sed -i 's#weston-start --#weston-start -- --continue-without-input#' ${D}/${sysconfdir}/init.d/weston
+    fi
+    if [ -e ${D}${systemd_system_unitdir}/weston.service ]; then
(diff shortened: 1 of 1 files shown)


recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -0,0 +1,10 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
+do_install:append:rpi() {
```
agherzan, line 3: What is the behaviour without a keyboard and without this option? This sounds like a oe-core change to me unless I'm missing something.

kraj: Currently, if you do not attached keyboard and mouse then weston will not start on rpi.
this is only needed for SOCs using DRM backends, so its really not something we can enumerate in OE-core.
there is `DEFAULTBACKEND` which can be set and then used to deduce if a machine is using drm backend but its not a strict requirement to mention one, since weston will pick one if not mentioned. So we do it here because with rpi we are sure of which backend is in use.

I have done something similar for meta-ti

agherzan: I was thinking of DEFAULTBACKEND indeed. But it makes sense now.

agherzan (approved):

agherzan:
@raj is this something introduced in later Weston versions?

kraj:
> @raj is this something introduced in later Weston versions?

yeah added in weston 9 onwards see [1]
[1] https://lists.freedesktop.org/archives/wayland-devel/2020-July/041555.html

agherzan:
8*

Thanks.

PR #967, opened 2021-12-05
recipes-bsp: Add support for gpio-shutdown

otski:
**- What I did**
Enable gpio-shutdown functionality, see https://github.com/raspberrypi/firmware/blob/master/boot/overlays/README for reference.

**- How I did it**
In short, this feature includes the gpio-shutdown.dtbo. If systemd init is enabled, no futher configuration is needed. However if (the default) SysV init is used, the GPIO input requires additional binding which is handled by recipes-bsp/gpio-shutdown/gpio-shutdown.bb:
1. gpio-shutdown sends key code 116 which is bound to KeyboardSignal and injected into the keymap in gpio-shutdown-keymap.sh. This init script is executed at runlevel S
2. SysV inittab needs instructions what to do when KeyboardSignal is received. This is implemented by bind_gpio_shutdown.tab which is deployed to /etc/inittab.d/

The feature is enabled by variable ENABLE_GPIO_SHUTDOWN. Additionally an alternative GPIO pin can be specified using variable GPIO_SHUTDOWN_PIN. Since the gpio-shutdown by default uses the same pins as the master I2C bus, rpi-config_git.bb includes a rudimentary sanity check for a configuration conflict.

I mainly tested this with core-image-full-cmdline, with both SysV init and systemd init on Raspberry Pi 1 Model B.

### DIFF
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -23,6 +23,7 @@ RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
     overlays/gpio-ir-tx.dtbo \
     overlays/gpio-key.dtbo \
     overlays/gpio-poweroff.dtbo \
+    overlays/gpio-shutdown.dtbo \
     overlays/hifiberry-amp.dtbo \
     overlays/hifiberry-dac.dtbo \
     overlays/hifiberry-dacplus.dtbo \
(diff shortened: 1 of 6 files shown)


agherzan (changes requested):
This looks pretty clean. We need to add some docs on the new variables.

otski:
> This looks pretty clean. We need to add some docs on the new variables.

Of course, added a section for the gpio-shutdown.

By the way, what's the policy on cherry-picking changes to other branches? I would like to have this on dunfell as well as it is the current LTS. Should this PR be merged, shall I make another PR for the dunfell branch? It seems it will also require some changes as dunfell has an older sysv init version that does not yet support inittab.d.

agherzan:
@otski the policy is that it first goes to master and once that is done, you can create an MR with the cherrypicks.

agherzan (approved):

PR #984, opened 2022-01-09
Add documentation to support RNDIS features

ajpenner:
**- What I did**
Added documentation to explain the build variables for RNDIS support

**- How I did it**
I modified the extra-build-config.md document to include a brief description of the variables. I have also added two new variables to allow the user to supply two MAC addresses for the host and device and add them to the cmdline.txt file

### DIFF
--- a/recipes-devtools/python/rpi-gpio_0.7.0.bb
+++ b/recipes-devtools/python/rpi-gpio_0.7.0.bb
@@ -5,7 +5,7 @@ LICENSE = "MIT"
 LIC_FILES_CHKSUM = "file://LICENCE.txt;md5=9b95630a648966b142f1a0dcea001cb7"
 
 PYPI_PACKAGE = "RPi.GPIO"
-inherit pypi distutils3
+inherit pypi setuptools3
 
 SRC_URI += "file://0001-Remove-nested-functions.patch"
 SRC_URI[md5sum] = "777617f9dea9a1680f9af43db0cf150e"


recipes-bsp/bootfiles/rpi-cmdline.bb
```diff
@@ -34,7 +34,13 @@ CMDLINE_DEBUG ?= ""
 # CMDLINE_RNDIS = "modules-load=dwc2,g_ether g_ether.host_addr=<some MAC 
 # address> g_ether.dev_addr=<some MAC address>"
 # if the MAC addresses are omitted, random values will be used
-CMDLINE_RNDIS ?= ""
+CMDLINE_RNDIS ?= '${@oe.utils.conditional("ENABLE_RNDIS", "1", "modules-load=dwc2,g_ether", "", d)}'
```
agherzan, line 37: `modules-load` is systemd-specific, right?

recipes-bsp/bootfiles/rpi-cmdline.bb
```diff
@@ -34,7 +34,13 @@ CMDLINE_DEBUG ?= ""
...
+
+TMP_HOST_MAC_ADDR = '${@oe.utils.conditional('RNDIS_HOST_MAC_ADDR', None, "", " g_ether.host_addr=" + str(d.getVar('RNDIS_HOST_MAC_ADDR')), d)}'
+CMDLINE_RNDIS_HOST_MAC_ADDR ?= '${@oe.utils.conditional("ENABLE_RNDIS", "1", TMP_HOST_MAC_ADDR, "", d)}'
```
agherzan, line 40: Would it be cleaner to use an anon function to handle this? It gets a bit tangled. Same for the set below.

recipes-bsp/bootfiles/rpi-cmdline.bb
agherzan, line 37: Isn't this dependent on the kernel configuration? module/built-in that is.

agherzan (changes requested):
Besides the comments I've raised, I would like to have the branch rebase and the history cleaned up (it would be nice to have one commit for docs and one for cmdline - but I wouldn't be too strict about it).

recipes-bsp/bootfiles/rpi-cmdline.bb
agherzan: In that case this should also have a check on systemd.

ajpenner: Yes it does, this requires that the OTG setting is also enabled. I will elaborate the dependencies to clarify this.

recipes-bsp/bootfiles/rpi-cmdline.bb
ajpenner: Great point

ajpenner: Done

recipes-bsp/bootfiles/rpi-cmdline.bb
ajpenner: Done

ajpenner:
I will realign my PR to make two as you suggest. I agree it is cleaner (but ran out of time today)

ajpenner:
Okay, I believe everything is in order for another review or accepting the PR.

Thanks!

PR #989, opened 2022-01-18
raspberrypi4-64: Switch to cortexa72 tune

pbrkr:
After commits ca50267ab568 & 03cebdd7ef92 in openembedded-core, the `cortexa72-crc` tune is no longer available and the `cortexa72` tune includes the crc extension by default. Update the raspberrypi4-64 machine config to handle these changes.


kraj (approved):
lgtm

PR #990, opened 2022-01-23
rpi-ssh: Add configuration option to enable ssh on first boot

ajpenner:
Signed-off-by: Andrew Penner <andrew.penner@protonmail.com>

**- What I did**
I added a new configuration option ENABLE_SSH to create the empty ssh file in /boot so the system will be able to enable ssh on first boot.

**- How I did it**
I created a new recipe in recipese-bsp/bootfiles that reads conf/local.conf for ENABLE_SSH. If ENABLE_SSH is set to "1" we touch the file /boot/ssh which is added to the final image.

### DIFF
--- a/docs/extra-build-config.md
+++ b/docs/extra-build-config.md
@@ -278,6 +278,14 @@ the header extension should set the following in local.conf:
 
     ENABLE_DWC2_HOST = "1"
 
+## Enable SSH support
+
+By default the raspberry pi's disable ssh. By setting the following variable
+we create the empty file "ssh" in the /boot/ directory which is used to enable
+the ssh capabilities on first boot. Note: This file will be deleted when used.
+
+    ENABLE_SSH = "1"
+
 ## Enable Openlabs 802.15.4 radio module
 
 When using device tree kernels, set this variable to enable the 802.15.4 hat:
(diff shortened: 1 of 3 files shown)


docs/extra-build-config.md
```diff
@@ -278,6 +278,14 @@ the header extension should set the following in local.conf:
 
     ENABLE_DWC2_HOST = "1"
 
+## Enable SSH support
+
+By default the raspberry pi's disable ssh. By setting the following variable
+we create the empty file "ssh" in the /boot/ directory which is used to enable
```
agherzan, line 284: What "consumes" this `ssh` file?

agherzan (changes requested):

docs/extra-build-config.md
ajpenner: Yep, good question. It turns out (as you likely know) nothing.

agherzan: RIght. So it must be some mechanism from another distro. Was it Raspbian?

ajpenner: Yeah, that is where I got the idea from. Following their lead I added it to my Yocto images. I did not actually double check that it was necessary.

PR #983, opened 2022-01-07
Add picamera support

bhargavthriler:
python3-picamera : recipe for python picamera library.
python3-colorzero: recipe for python colorzero library, required by picamera
picamera-libs    : recipe raspberry pi firmware which are required by picamera.

Closes: #959
Signed-off-by: bhargavthriler <bhargavthriler@gmail.com>

**- What I did**
Created recipe for python picamera library. Along with it created 2 more, python colorzero and picamera-libs.

PR #994, opened 2022-01-26
picamera-libs,python3-picamera: Limit visibility to rpi machines

kraj:
https://github.com/agherzan/meta-raspberrypi/pull/983 introduced these
recipes however these are not generic packages that can be used on non-rpi machines
therefore mark them so. It helps in using this SOC layer with other SOC
layers in a distro

### DIFF
--- a/recipes-multimedia/picamera-libs/picamera-libs.bb
+++ b/recipes-multimedia/picamera-libs/picamera-libs.bb
@@ -21,3 +21,6 @@ INHIBIT_PACKAGE_STRIP = "1"
 INHIBIT_SYSROOT_STRIP = "1"
 SOLIBS = ".so"
 FILES_SOLIBSDEV = ""
+
+COMPATIBLE_HOST = "null"
+COMPATIBLE_HOST:rpi:libc-glibc = "(arm.*)-linux"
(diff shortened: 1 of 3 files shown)


recipes-multimedia/picamera-libs/picamera-libs.bb
```diff
@@ -21,3 +21,6 @@ INHIBIT_PACKAGE_STRIP = "1"
 INHIBIT_SYSROOT_STRIP = "1"
 SOLIBS = ".so"
 FILES_SOLIBSDEV = ""
+
+COMPATIBLE_HOST = "null"
+COMPATIBLE_HOST:rpi = "(.*)"
```
agherzan, line 26: Isn't this supposed to be COMPATIBLE_MACHINE?

kraj: Both work. Host is better since it may not mark the recipe machine specific when it is compiling

agherzan (approved):

recipes-multimedia/picamera-libs/picamera-libs.bb
agherzan: But I guess they should be machine specific in this case, shouldn't they?

kraj: It won’t matter but if you prefer then I would make the change

kraj: hang on this recipe needs more love. Let me finish all and upload PR

kraj:
@agherzan  this should be better now. Please take a look.

agherzan (approved):

recipes-multimedia/picamera-libs/picamera-libs.bb
agherzan: Looks good. Thanks and merged.

PR #1000, opened 2022-02-03
rpi-cmdline: Allow user to configure the raspberry pi in gadget mode

ajpenner:
Signed-off-by: Andrew Penner <andrew.penner@protonmail.com>

**- What I did**
I added several new configuration options to allow users to enable different gadget modes. These include:
g_serial, g_ether, g_cdc, and c_multi.

**- How I did it**
I modified the existing cmdline recipe to use new configuration variables to modify the text in cmdline.txt so the booted device will load the correct modules that enable the Raspberry Pi to operate as a gadget.

### DIFF
--- a/recipes-bsp/bootfiles/rpi-cmdline.bb
+++ b/recipes-bsp/bootfiles/rpi-cmdline.bb
@@ -29,12 +29,65 @@ CMDLINE_LOGO ?= '${@oe.utils.conditional("DISABLE_RPI_BOOT_LOGO", "1", "logo.nol
 # to enable kernel debugging.
 CMDLINE_DEBUG ?= ""
 
-# Add RNDIS capabilities (must be after rootwait)
-# example: 
-# CMDLINE_RNDIS = "modules-load=dwc2,g_ether g_ether.host_addr=<some MAC 
-# address> g_ether.dev_addr=<some MAC address>"
+# Add gadget capabilities (must be after rootwait)
 # if the MAC addresses are omitted, random values will be used
-CMDLINE_RNDIS ?= ""
+def setup_gadget_mac(d, gadget_type):
+    string = ""
+    if d.getVar('GADGET_HOST_MAC_ADDR'):
+        string += " " + gadget_type + ".host_addr="\
+          + d.getVar('GADGET_HOST_MAC_ADDR')
+    if d.getVar('GADGET_DEV_MAC_ADDR'):
+        string += " " + gadget_type + ".dev_addr="\
+          + d.getVar('GADGET_DEV_MAC_ADDR')
+    return string
+
+def setup_mass_storage(d, gadget_type):
+    string = ""
+    if d.getVar('GADGET_MASS_STORAGE_NAME'):
+        string += " " + gadget_type + ".file="\
+          + d.getVar('GADGET_MASS_STORAGE_NAME')
+    return string
+
+# Setup the gadget configuration
+def setup_gadget_mode(d):
+    string = ""
+    if not bb.utils.contains('DISTRO_FEATURES', 'systemd', True, False, d)\
+        or d.getVar('ENABLE_DWC2_PERIPHERAL') != "1":
+            return string
+
+    if d.getVar('ENABLE_ETHER_GADGET') == "1":
+        string += "modules-load=dwc2,g_ether"
+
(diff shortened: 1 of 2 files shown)


recipes-bsp/bootfiles/rpi-cmdline.bb
```diff
@@ -11,7 +11,17 @@ CMDLINE_DWC_OTG ?= "dwc_otg.lpm_enable=0"
 CMDLINE_ROOT_FSTYPE ?= "rootfstype=ext4"
 CMDLINE_ROOTFS ?= "root=/dev/mmcblk0p2 ${CMDLINE_ROOT_FSTYPE} rootwait"
 
-CMDLINE_SERIAL ?= "${@oe.utils.conditional("ENABLE_UART", "1", "console=serial0,115200", "", d)}"
+# revert the original serial setup
+def serial_comms(d):
```
agherzan, line 15: I'm not sure how this is relevant for this commit. Maybe I'm missing something but this is already flexible by using the `CMDLINE_SERIAL` variable. What is your use case here?

ajpenner: More experimentation shows that I do not actually need the UART enabled at all for a serial connection. This change will be backed out.

Originally I was under the impression I needed the string " console=tty1" appended to the CMDLIE_SERIAL, and I thought this was the cleanest way to do that. I am open to suggestions if there is a better approach to that (for the next time this might come up)

agherzan: @ajpenner can you please clean up and include the documentation here too? Closing https://github.com/agherzan/meta-raspberrypi/pull/1001 assuming that you pull include it as part of this PR.

ajpenner: I have reverted the CMDLINE_SERIAL change since it was not needed. I have added the documentation here as well.

ajpenner:
Is there something else I need to change?

recipes-bsp/bootfiles/rpi-cmdline.bb
```diff
@@ -29,12 +29,65 @@ CMDLINE_LOGO ?= '${@oe.utils.conditional("DISABLE_RPI_BOOT_LOGO", "1", "logo.nol
...
+    if d.getVar('GADGET_MASS_STORAGE_NAME'):
+        string += " " + gadget_type + ".file="\
+          + d.getVar('GADGET_MASS_STORAGE_NAME')
+    return string
+
+# Setup the gadget configuration
+def setup_gadget_mode(d):
+    string = ""
+    if not bb.utils.contains('DISTRO_FEATURES', 'systemd', True, False, d)\
+        or d.getVar('ENABLE_DWC2_PERIPHERAL') != "1":
```
agherzan, line 55: Could we add an incompatibility warning/error? Otherwise, configuration might get ignored silently.

recipes-bsp/bootfiles/rpi-cmdline.bb
```diff
@@ -29,12 +29,65 @@ CMDLINE_LOGO ?= '${@oe.utils.conditional("DISABLE_RPI_BOOT_LOGO", "1", "logo.nol
...
+    string = ""
+    if d.getVar('GADGET_MASS_STORAGE_NAME'):
+        string += " " + gadget_type + ".file="\
+          + d.getVar('GADGET_MASS_STORAGE_NAME')
+    return string
+
+# Setup the gadget configuration
+def setup_gadget_mode(d):
+    string = ""
+    if not bb.utils.contains('DISTRO_FEATURES', 'systemd', True, False, d)\
```
agherzan, line 54: What's the dependency on systemd here?

ajpenner: This was requested because the overlays require systemd for the OTG module to load. If this is not the case, let me know and I will remove it.

agherzan: I don't see how that is dependent on the init system chosen. First of all, because you can load modules with any init system and secondly because you can have the module built-in. Isn't that the case?

ajpenner:
Some code here was found during paid employment, so they may own the copyright not me

PR #916, opened 2021-09-30
FSFE-REUSE Booster Recomended Updates

lnceballosz:
Hello,

By way of introduction, I am Lina Ceballos from the Free Software Foundation Europe, and the REUSE Booster program. After a quick check on your repository, I would like to propose some updates regarding your copyright and licensing information. As you may be aware, the [REUSE](https://reuse.software/) specifications intend to make licensing easier with best practices to display legal information through comment headers on source files that can be easily human and machine readable. You can check this screencast:  
https://download.fsfe.org/videos/reuse/screencasts/reuse-tool.gif

I would also like to make some remarks to be considered to make your full project REUSE compliant.

Feel completely free to adopt these specifications by merging this pull request.

### REUSE Features:

1. SPDX copyright and license comment headers for all relevant files (please be aware that to serve as example I added headers to most of the files. However, some files might be missing this feature).
2. LICENSES directory in the root of the project with the licenses used on the repository:
- I included in this directory the file that contained the MIT license text. Since I added a header to the file `.gitignore `that originally is not copyrightable, under the CC0 1.0, this license text is also included in the directory.
- I also noticed that your repository has a custom license **"Broadcom-RPi"** I didn't license any file under this custom license, since I am not sure what the purpose of this license is and/or what files are under the same. Please make sure to add this information, as well as the license text to the` LICENSES` directory. You can find more information about this [here](https://reuse.software/faq/#custom-license).

###  Files missing copyright and licensing information:

I added the comment headers with copyright and license information to some of the files that lacked that information.

I provided a more detailed license information to` README` file as well

There is an image file in the` img/ `directory. Since I am not sure who the copyright holder is and under what license this file is, I didn't create any license file for this image. You can find further information on how to add headers to binary or uncommentable files [here](https://reuse.software/tutorial/#binary-and-uncommentable-files).

Please also double check if the personal information in the headers is correct and consistent, in the case of several copyright holders please update that information in the headers.

**Note:** Your project has some large directories with patch files e.g `recipes-graphics/userland/files/`, to mention one. From my understanding, these are contributions made to your project in the past. Are these contributions already implemented in your project? You can keep these files for historical purposes, but if these contributions are already part of your project, my advice would be to delete these files. If you want to keep them, you will also have to determine under what license these files will be. Remember that the fundamental idea of REUSE is that all your files are marked with license and copyright information.

We can also do a follow up of this PR and further REUSE specification with a call. Feel free to reach me out.

Hope that helps and thank you very much for the amazing job!

PR #1005, opened 2022-02-09
ci: Introduce initial GitHub workflows

agherzan:
This pull request provides the initial workflows for bringing back the CI validations and improving them with additional checks:
1. Build all the currently supported boards with `poky/systemd` configuration on `core-image-base`. As I get (happy to hear from you if you want to help) more runners I can easily extend it for more configuration (`uboot/firmware`, `userland/VC4` etc.).
2. Fix small nits for having the layer pass the yocto layer checker script.
3. Check the layer with the yocto layer checker script - we can now apply for yocto compatible.
4. Check for DCO - SoB needs to match the Author.
5. Check for reuse - currently, the result of this job is not affecting the CI overall outcome as the layer is not yet reuse compliant. That is already WIP - https://github.com/agherzan/meta-raspberrypi/pull/916.


.github/workflows/docker-images/yocto-builder/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM ubuntu:20.04
+
+ARG DEBIAN_FRONTEND="noninteractive"
```
zyga, line 3: Why ARG and not ENV?

agherzan: Because there is no need to pass these after building the image.

.github/workflows/docker-images/yocto-builder/entrypoint-build.sh
```diff
@@ -0,0 +1,46 @@
...
+[ -z "$SSTATE_DIR" ] || echo SSTATE_DIR = \"$SSTATE_DIR\" >> conf/local.conf
+[ -z "$DL_DIR" ] || echo DL_DIR = \"$DL_DIR\" >> conf/local.conf
+echo BB_NUMBER_THREADS = \"6\" >> conf/local.conf
+echo PARALLEL_MAKE = \"-j 6\" >> conf/local.conf
+
+bitbake-layers add-layer "$META_RASPBERRYPI_PATH"
+
+echo "\nlocal.conf:\n--------------"
+cat conf/local.conf
+echo "\nbblayers.conf:\n--------------"
```
agherzan, line 43: FIXME

.github/workflows/docker-images/yocto-builder/entrypoint-build.sh
```diff
@@ -0,0 +1,55 @@
...
+
+REPOS=" \
+    git://git.yoctoproject.org/poky.git \
+"
+for repo in $REPOS; do
+    log "Cloning $repo on branch $BASE_REF..."
+    git clone --depth 1 --branch "$BASE_REF" "$repo"
+done
+
+# shellcheck disable=SC1091,SC2240
```
zyga, line 34: I'm curious if `shellcheck -x` is actually able to follow this one?

.github/workflows/yocto-builds.yml
```diff
@@ -0,0 +1,65 @@
...
+          - raspberrypi0-wifi
+          - raspberrypi2
+          - raspberrypi3-64
+          - raspberrypi3
+          - raspberrypi4-64
+          - raspberrypi4
+          - raspberrypi-cm3
+          - raspberrypi-cm
+        image: [core-image-base]
+        dl-dir: [/var/lib/ci/yocto/downloads]
```
zyga, line 29: Nitpick: I'd put those loop invariant variables to the regular variables section.

.github/workflows/yocto-builds.yml
```diff
@@ -0,0 +1,65 @@
...
+        distro: [poky]
+    runs-on: [self-hosted, Linux]
+    name: ${{ matrix.machine }}/${{ matrix.image }}/poky/systemd
+    steps:
+      - name: Checkout the code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Build a temporary yocto-builder image
+        uses: ./.github/actions/docker-build
```
zyga, line 40: This feels wasteful.

Can you pull from a registry instead?

.github/workflows/docker-images/yocto-builder/entrypoint-build.sh
```diff
@@ -0,0 +1,55 @@
...
+. ./poky/oe-init-build-env build
+
+# Build configuration
+printf "\n# ------ ci ------\n" >> conf/local.conf
+[ -z "$SSTATE_DIR" ] || echo SSTATE_DIR = \""$SSTATE_DIR"\" >> conf/local.conf
+[ -z "$DL_DIR" ] || echo DL_DIR = \""$DL_DIR"\" >> conf/local.conf
+[ -z "$DISTRO" ] || echo DISTRO = \""$DISTRO"\" >> conf/local.conf
+{
+    echo BB_NUMBER_THREADS = \"6\"
```
zyga, line 43: Nitpick, cat and here-doc might make this easier to quote and would remote the need for subshell.
```
cat <<STUFF >>conf/local.conf
PARALLEL_MAKE = "-j 6"
...
STUFF
```

.github/workflows/docker-images/yocto-builder/README.md
```diff
@@ -0,0 +1,16 @@
+<!--
+SPDX-FileCopyrightText: Andrei Gherzan <andrei.gherzan@huawei.com>
+
+SPDX-License-Identifier: MIT
+-->
+
+# Docker image for builds
+
+This defines the docker image for running Yocto/OE based operations/builds. It
+privides multiples scripts for driving different operations.
```
Stefan-Schmidt, line 10: Typo, provides multiple

.github/workflows/docker-images/yocto-builder/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN eatmydata apt-get install -qq -y locales \
+	&& echo "en_US.UTF-8 UTF-8" > /etc/locale.gen \
+	&& locale-gen
+ENV LANG en_US.utf8
+
+RUN eatmydata apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# Have bash as shell
+RUN echo "dash dash/sh boolean false" | debconf-set-selections \
+ && DEBIAN_FRONTEND=noninteractive dpkg-reconfigure dash
```
esben, line 30: No need to set `DEBIAN_FRONTEND` again here.

esben (dismissed):

.github/workflows/docker-images/yocto-builder/Dockerfile
agherzan: Cheers! Fixed it.

.github/workflows/docker-images/yocto-builder/entrypoint-build.sh
agherzan: I'm not sure I understand what you mean here. The `disable` above instructs `shellcheck` to not try to follow the next line. I might misunderstand your question though.

.github/workflows/docker-images/yocto-builder/entrypoint-build.sh
agherzan: I like that. Fixed.

.github/workflows/yocto-builds.yml
agherzan: Sounds fair. Fixed.

.github/workflows/yocto-builds.yml
agherzan: I agree with you that this looks a bit cumbersome and maybe useless. The advantage of having this as such is the ability to modify the docker images (Dockerfiles) along with the extra layer metadata as part of the same PR (repo). I'll test drive this for a while and based on how it behaves I might change to a separate repo for the docker files and use a registry which would make this cleaner but with the disadvantage of having to use multiple repos/PRs.

esben: Another solution would be to keep the Dockerfile in this repo, but push to a registry, and then do a pull from that and use that as cache, so you don't have to waste time on rebuilding unless actually needed.

kraj:
the patches seems incremental fixes. Can you squash them into changelogs that are rewiewable units ?

agherzan:
> the patches seems incremental fixes. Can you squash them into changelogs that are rewiewable units ?

@kraj they should be. Could you give me an example?

.github/workflows/yocto-builds.yml
esben: I normally restrict the push to registry to only happen for official branches.
Depending on requirements, I bump a version number in the image tag, so old image versions are maintained.

For testing purposes, I either let that require a manual trigger, or push to a "test" image, which is then recycled for each build.

agherzan: Using one test image would have a race condition for multiple PRs. Also, you need to make sure multiple PRs don't use images built for another PR. So it would need per PR images. And even if you do, there is no guarantee that the current PR  image is synced with the current iteration for the code. So you would need to rebuild anyway (with existing layers in the repository so not from scratch as to how I've done it here). The issue though is when you have jobs running for multiple iterations of the same MR due to the fact that stopping jobs in GitHub is not a baked functionality - and it is dealt with usually but an additional pipeline that works on start of PRs.

All these are much easier if you deal with local images where you assume one worker per docker storage. The image cleanup it's a disk optimization because I couldn't find an easy way to run a job on merge (it becomes a push where you don't have the reference of the originating PR). If I figure that out, I could reuse build images locally until I merge when I deal with cleanups.

kraj (approved):

PR #1038, opened 2022-03-19
Added a ENABLE_PWM config setting

vermaete:


kraj (dismissed):
lgtm

docs/extra-build-config.md
```diff
@@ -290,18 +290,24 @@ See: <https://openlabs.co/OSHW/Raspberry-Pi-802.15.4-radio>
...
-	ENABLE_DUAL_CAN = "1"
+    ENABLE_DUAL_CAN = "1"
 
 Some modules may require setting the frequency of the crystal oscillator used on the particular board. The frequency is usually marked on the package of the crystal. By default, it is set to 16 MHz. To change that to 8 MHz, the following variable also has to be set:
 
     CAN_OSCILLATOR="8000000"
 
+## Enable PWM
+
+In oreder to use PWM, set the following variable:
```
agherzan, line 307: A small typo here.

agherzan (changes requested):

docs/extra-build-config.md
vermaete: Should be fixed.  Sorry.

agherzan: Thanks!

agherzan:
@vermaete can you please squash the commits accordingly? There is no need to clutter the history.

vermaete:
Hi,
Is there still something missing from me?

kraj (approved):
lgtm now

agherzan (changes requested):
This looks good Jan. Some small tweaks if you don't mind:
1. It is standard to avoid a full stop in the git subject
2. Git log should be a description of the change not a call to the maintainers :)
3. Changes that are out of the scope of the change should be split into separate commits - for example, there are some reformatting changes.

recipes-bsp/bootfiles/rpi-config_git.bb
```diff
@@ -255,6 +255,12 @@ do_deploy() {
         echo "dtoverlay=mcp2515-can0,oscillator=${CAN_OSCILLATOR},interrupt=25" >>$CONFIG
     fi
 
+    # PWM support
+    if [ "${ENABLE_PWM}" = "1" ]; then
+        echo "# Enable PWM" >>$CONFIG
+        echo "dtoverlay=pwm,pin=18,func=2" >>$CONFIG
```
agherzan, line 261: Might not be a good idea to hardcode this. If I remember this correctly, it conflicts with the i2c audio pins. Also, `18` is the default anyway, so we might just drop it and address a configuration variable later. Same for `func`.

agherzan: That sounds right.

PR #1041, opened 2022-03-27
u-boot: always set fdt_addr with firmware-provided FDT address

matthiasklein:
The Raspberry firmware prepares the FDT blob in memory at an address that
depends on both the memory size and the blob size.
For details see description in the patch itself.

The patch is not yet upstream, but has already been submitted by the original
author: https://lore.kernel.org/all/20210512123945.25649-1-m.salvini@koansoftware.com/

For me, the described problem occurs in conjunction with RAUC, which uses the
following U-Boot script: https://github.com/rauc/meta-rauc-community/blob/master/meta-rauc-raspberrypi/recipes-bsp/rpi-u-boot-scr/files/boot.cmd.in

As soon as I add an overlay the fdt_addr changes and the system does not boot
anymore.

### DIFF
--- a/recipes-bsp/u-boot/u-boot_%.bbappend
+++ b/recipes-bsp/u-boot/u-boot_%.bbappend
@@ -4,6 +4,8 @@ SRC_URI:append:rpi = " \
     file://fw_env.config \
 "
 
+SRC_URI:append:rpi = " file://0001-rpi-always-set-fdt_addr-with-firmware-provided-FDT-address.patch"
+
 # special fix for raspberrypi-cm3
 SRC_URI:append:raspberrypi-cm3 = " file://0001-dm-core-Move-ofdata_to_platdata-call-earlier.patch"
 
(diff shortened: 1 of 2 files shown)


recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -4,6 +4,8 @@ SRC_URI:append:rpi = " \
     file://fw_env.config \
 "
 
+SRC_URI:append = " file://0001-rpi-always-set-fdt_addr-with-firmware-provided-FDT-address.patch"
```
agherzan, line 7: I'm wondering if it would be a bit better to use the `rpi` override. The advantage I see is that other setups would not break if the patch doesn't apply anymore forcing only the Raspberry Pi machines to update the patch. Given that this is a Raspberry Pi specific patch, it would make sense.

agherzan (dismissed):

matthiasklein:
I have added an rpi override accordingly.

agherzan (approved):
Cheers, @matthiasklein!

PR #1046, opened 2022-04-04
docs: link to latest documentation of kas

vermaete:
Version 1.0 of kas is not the latest anymore.

Signed-off-by: Jan Vermaete <jan.vermaete@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ local_conf_header:
 To configure the machine, you have to update the `machine` variable.
 And the same for the `distro`.
 
-For further information, you can read more at <https://kas.readthedocs.io/en/1.0/index.html>
+For further information, you can read more at <https://kas.readthedocs.io/en/latest/index.html>
 
 ## Contributing
 


README.md
```diff
@@ -68,7 +68,7 @@ local_conf_header:
 To configure the machine, you have to update the `machine` variable.
 And the same for the `distro`.
 
-For further information, you can read more at <https://kas.readthedocs.io/en/1.0/index.html>
+For further information, you can read more at <https://kas.readthedocs.io/en/latest/index.html>
```
agherzan, line 71: I'm not sure if this is a good idea or not given that we use version 8 in the kas configuration file. We should stick to whatever that configuration uses. What is the problem that you are trying to solve here?

agherzan (changes requested):

README.md
vermaete: Hi,
Sorry for the late answer.

My thinking was more that the Kas people probably think their latest release is the greatest.
As Yocto people do :)

The 1.0 release of Kas is from 21 May 2019.  That's about the time of the Warrior release of Yocto.

It was just a quick online editing.  Not something to solve, just keep the documentation in line with the version most people will download and install first.
And a bit of promotion of Kas, too.

br

agherzan: Fair. I'm fine with that.

Issue #1054, opened 2022-04-24
Kernel modules not loaded during boot


bhargavthriler:
**Description**
Observed that some kernel modules are not being loaded in the latest kernel `5.15.34-v7`.

So I have built a `core-image-base` with the tip of the master (`0135a02`) and while trying access the camera using Picamera got some errors. The errors mainly complain about mmal drivers not present.
```
root@raspberrypi3:~# python3
Python 3.10.4 (main, Mar 23 2022, 20:25:24) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from picamera import PiCamera
>>> camera = PiCamera()
mmal: mmal_vc_shm_init: could not initialize vc shared memory service
mmal: mmal_vc_component_create: failed to initialise shm for 'vc.camera_info' (7:EIO)
mmal: mmal_component_create_core: could not create component 'vc.camera_info' (7)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.10/site-packages/picamera/camera.py", line 408, in __init__
    self._init_revision(options)
  File "/usr/lib/python3.10/site-packages/picamera/camera.py", line 480, in _init_revision
    with mo.MMALCameraInfo() as camera_info:
  File "/usr/lib/python3.10/site-packages/picamera/mmalobj.py", line 2425, in __init__
    super(MMALCameraInfo, self).__init__()
  File "/usr/lib/python3.10/site-packages/picamera/mmalobj.py", line 696, in __init__
    mmal_check(
  File "/usr/lib/python3.10/site-packages/picamera/exc.py", line 184, in mmal_check
    raise PiCameraMMALError(status, prefix)
picamera.exc.PiCameraMMALError: Failed to create MMAL component b'vc.camera_info': I/O error
>>> 
>>> 
root@raspberrypi3:~# 
```

After digging through my system found an older build (don't know why didn't delete it but thankfully it gave some insight into the issue), I tried booting that image and everything seems to working fine.

So I checked out to the commit which the older build was using (`63a3d8cb17c5d1affe8f2848f45fcc6a706f9412`), and the camera worked fine(though I had to make few changes, which are not significant for this issue). While analyzing the bootlogs found that the latest build (`0135a02`) doesn't load all the drivers.

[...]

PR #1056, opened 2022-04-27
kmod: Enable xz compression

agherzan:
The RaspberryPi default Linux kernel configuration uses xz for
compressing kernel modules[1]. This was introduced in 5.15. We need the
respective support in kmod too.

[1]https://github.com/raspberrypi/linux/pull/4991

Fixes: https://github.com/agherzan/meta-raspberrypi/issues/1054

### DIFF
--- a/recipes-kernel/kmod/kmod_%.bbappend
+++ b/recipes-kernel/kmod/kmod_%.bbappend
@@ -1,0 +1,5 @@
+# SPDX-FileCopyrightText: Andrei Gherzan <andrei@gherzan.com>
+#
+# SPDX-License-Identifier: MIT
+
+PACKAGECONFIG:append:rpi = " xz"


recipes-kernel/kmod/kmod_%.bbappend
```diff
@@ -0,0 +1,5 @@
+# SPDX-FileCopyrightText: Andrei Gherzan <andrei@gherzan.com>
+#
+# SPDX-License-Identifier: MIT
+
+PACKAGECONFIG:append:rpi = " xz"
```
kraj, line 5: this will enable it for target but the real problem is `kmod-native` so we need 
```
PACKAGECONFIG:append = " xz"
```

but this will change the generic options and the layer will lose yocto compliance. So best is to enable it in main recipe. So please lobby for the [patch](https://lists.openembedded.org/g/openembedded-core/message/164932) I have sent :)

kraj: even enabling `xz` for  kmod-native is not resolving this :( so something more is amiss

agherzan: I was just about to do this. Thanks for handling it.

kraj:
oe-core commit in on its way into master so we could have avoided adding this by waiting a little

agherzan:
@kraj I misread a comment. I'll revert the commit.

PR #1057, opened 2022-04-29
rng-tools: disable libjitterentropy due to cpu usage

wesQ3:
Since updating to kirkstone from hardknott, after boot rngd maxes out my rpi4's processor for minutes initializing JITTER. The sustained CPU usage was triggering my resource monitoring alerts. Changing config to disable jitter with `-x jitter` stops the initialization process and uses just the pi's hardware rng source. Since that solved the problem I disabled building rng-tools with libjitterentropy enabled.

If this is a problem for other users devices I thought this simple change could be helpful.

**- What I did**
Disable libjitterentropy in rng-tools.

**- How I did it**
Set `PACKAGECONFIG` to override the default poky setting which is `libjitterentropy`.

### DIFF
--- a/recipes-support/rng-tools/rng-tools_%.bbappend
+++ b/recipes-support/rng-tools/rng-tools_%.bbappend
@@ -1,0 +1,5 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
+# disable libjitterentropy because of sustained cpu use after boot
+# instead we depend on /dev/hwrng via config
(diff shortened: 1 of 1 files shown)


kraj (changes requested):
The fix is fine but I think it should be in core layer since even qemu machines are affected by this and beaglebone also have problem

wesQ3:
Closing this PR as [discussion is now on the oe-core mailing list](https://lists.openembedded.org/g/openembedded-core/topic/90845997).

PR #1074, opened 2022-07-27
Various recipe updates

agherzan:
This MR is to be kept as a draft until we figure out:

1. https://github.com/RPi-Distro/firmware-nonfree/issues/28
2. ~~https://github.com/RPi-Distro/firmware-nonfree/issues/29~~ - we have surfaced this under a license flag


recipes-kernel/linux/linux-raspberrypi_5.10.bb
```diff
@@ -2,8 +2,8 @@ LINUX_VERSION ?= "5.10.110"
 LINUX_RPI_BRANCH ?= "rpi-5.10.y"
 LINUX_RPI_KMETA_BRANCH ?= "yocto-5.10"
 
-SRCREV_machine = "89c0af71c9cf157a865afb526e9ebc21aadd531b"
-SRCREV_meta = "e1979ceb171bc91ef2cb71cfcde548a101dab687"
+SRCREV_machine = "427c6bd8835e197693e9b4aedbe45c2c3c84cdce"
+SRCREV_meta = "96ea2660bb97e15f48f4885b9e436f24c3606bd9"
```
kraj, line 7: I think we also need to bump LINUX_VERSION to 5.10.133 or something

kraj (changes requested):

recipes-kernel/linux/linux-raspberrypi_5.10.bb
agherzan: I don't think so unless I'm missing something obvious https://github.com/raspberrypi/linux/blob/rpi-5.10.y/Makefile#L4

kraj: hmm OK. I was looking at kernel.org and there is 5.10.133 latest on 5.10 series. but maybe raspberrypi-linux has stopped maintaining 5.10 ?

agherzan: That could be. Or it didn't catch up.

kraj (approved):
lgtm

kraj (approved):
lgtm

PR #1080, opened 2022-09-01
rpi-cmdline: do_compile: Use pure Python syntax to get `CMDLINE`

lluiscampos:
Otherwise the shell snippet fails with `bad syntax` when `CMDLINE`
contains special characters like `${...}` (useful to insert a U-Boot
variable in `cmdline.txt`).

**- What I did**

Modify `do_compile` so that special characters `${...}` are accepted for `CMDLINE`.

**- How I did it**

Getting `CMDLINE` variable through Python with `d.getVar()`

### DIFF
--- a/recipes-bsp/bootfiles/rpi-cmdline.bb
+++ b/recipes-bsp/bootfiles/rpi-cmdline.bb
@@ -62,7 +62,7 @@ CMDLINE = " \
     "
 
 do_compile() {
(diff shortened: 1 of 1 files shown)


lluiscampos:
@agherzan Thanks for your review and approval. What is the process to port patches to stable branches? I would like to bring this one to `kirkstone`.

agherzan:
@lluiscampos As soon as it lands in `master`, just create another MR cherry-picking it for `kirkstone`. It'd be the fastest way.

PR #1082, opened 2022-09-07
Kirkstone

marvinfrankeengram:
**- What I did**
Fixed a gstreamer issue occuring when 

**- How I did it**

### DIFF
--- a/recipes-bsp/bootfiles/rpi-cmdline.bb
+++ b/recipes-bsp/bootfiles/rpi-cmdline.bb
@@ -62,7 +62,7 @@ CMDLINE = " \
     "
 
 do_compile() {
-    echo "${@' '.join('${CMDLINE}'.split())}" > "${WORKDIR}/cmdline.txt"
+    echo "${@' '.join(d.getVar('CMDLINE').split())}" > "${WORKDIR}/cmdline.txt"
 }
 
 do_deploy() {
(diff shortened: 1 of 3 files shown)


agherzan:
@marvinfrankeengram Isn't `master` affected by this?

agherzan:
Also, I'd rather use upstream backported patches as opposed to ones that are taken from various online places. Was this issue fixed upstream?

agherzan:
Could you point me to the patches that Raspberry Pi OS is applying to fix this?

marvinfrankeengram:
Raspberry Pi OS bullseye uses 18.4, bookworm uses 20.3. It seems they don't include these fixes yet. The fixes are backported to the 1.20 Branch, not being part of any release yet.
https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/2793
https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/1094

agherzan:
@marvinfrankeengram so in that case why don't we backport those specific patches? Also, they don't look very specific to Raspberry Pi. Wouldn't it make more sense to push this to oe-core?

marvinfrankeengram:
They are the same. This patch is a combination of the second merge request from the issue and 2793.
I added this patch to the raspberry layer, because the bug seems to only occur on pi's. Tried to reproduce it on a machine with an intel processor, but it worked without the patch.

agherzan:
@marvinfrankeengram this is not because it is RPi specific. It is because you haven't tested it with the right colorspace. So I'd encourage you to take this to oe-core. And while doing so make sure you git-format the patches directly from the upstream repository so the history is preserved and easier to trace.

PR #1083, opened 2022-09-07
Multiconfig support (Raspberry Pi 3 + 4)

agherzan:
This is an initial proposal for being able to build images that boot on multiple targets. The current implementation adds support for Raspberry Pi 3 and 4.


conf/machine/include/rpi-base.inc
```diff
@@ -150,3 +150,20 @@ do_image_wic[recrdeps] = "do_build"
 # The kernel image is installed into the FAT32 boot partition and does not need
 # to also be installed into the rootfs.
 RDEPENDS:${KERNEL_PACKAGE_NAME}-base = ""
+
+RASPBERRYPI_3_FIRMWARE_PACKAGES += "\
```
rzr, line 154: For easier maintenance, i would have split those variables in separate include files (one per device type)
and include them all in the rpi-base

conf/multiconfig/raspberrypi3.conf
```diff
@@ -0,0 +1,7 @@
+# SPDX-FileCopyrightText: Andrei Gherzan <andrei.gherzan@huawei.com>
+#
+# SPDX-License-Identifier: MIT
+
+MACHINE = "raspberrypi3"
+
+TMPDIR = "${TOPDIR}/tmp-multi-raspberrypi3"
```
rzr, line 7: could reuse ${MACHINE} to align others

docs/multiconfig.md
```diff
@@ -0,0 +1,40 @@
...
+    bitbake mc:raspberrypi:core-image-base
+
+The final image will be available in:
+
+    <BUILDDIR>/tmp-multi-raspberrypi/deploy/images/raspberrypi4/
+
+This image will boot and provide the support for all the machines listed below.
+
+## Supported Machines for Multiconfig Builds
+
```
rzr, line 38: Is anything blocking for pi2 (and pi0?)

conf/machine/include/rpi-base.inc
agherzan: I will skip that for now as I don't want to introduce more confusion by further splitting the machine-specific confs/incs vs the confs that are applied for all machines.

docs/multiconfig.md
agherzan: Not technically but I haven't aimed at testing them at this point. I can pick it up later if nobody else does.

agherzan:
This won't work due to the complication of multiple kernel modules. I'll rethink this using multiple kernel recipes.

kraj:
Looks good approach to me. Although a bit stretch use of multiconfig but thats fine.

agherzan:
@kraj this would have worked fine if it wasn't for the modules complication. I am testing now a different approach - https://github.com/agherzan/meta-raspberrypi/commits/ag/raspberrypi-armv7

PR #1091, opened 2022-09-20
Follow-up REUSE specifications

lnceballosz:
Hi 

I'm Lina Ceballos from the [Free Software Foundation Europe,](https://fsfe.org/) and the REUSE Booster program. As promised, here it's a cleaner and more updated merge request which suggests the [REUSE specifications](https://reuse.software/spec/) that make licensing and copyright information unambiguous and perfectly human- and machine-readable [^0]. This MR is a follow-up from the [one](https://github.com/agherzan/meta-raspberrypi/pull/916) I closed some weeks ago.

Some important notes in this regard:

- Following our discussion I didn't include the year in the headers, as well as they should all be uniform now.
- I kept insignificant files such as .gitignore under MIT, in line with our earlier discussion, and documentation files are now under CC-BY-4.0.
- I have also created a[ dep5 file ](https://reuse.software/spec/#dep5)in the `.reuse` folder that includes patch and conf files, keeping in mind that creating a separate .license file for each of them is not ideal. However, if you prefer to proceed with the latter, please feel free to do so.

There are still **some files that I didn't touch**:
The imagine files `img/LF_17_02_Yocto-Badge-Update_Compatible_Final_Blank.png` and` img/balena.png,` since I don't know who the copyright holder is and under what license they are. Please feel free to add this information by creating a .license file for those two image files.

For the files:
`recipes-bsp/bootfiles/rpi-bootfiles.bb`
`recipes-graphics/vc-graphics/vc-graphics.inc`

Keeping in mind that they have a special license situation,  my suggestions here remain as follows:

You can keep the license MIT for those files since they are licensed under MIT although the components they build are under the custom license: Broadcom-RPi license.

In this regard, I noticed that the text of the custom license makes it a proprietary license. This apart, to be more precise:

_"This software may only be used for the purposes of developing for, running, or using a Raspberry Pi device."_

The custom license still has to be part of the `LICENSE` directory, you can add it following [these steps](https://reuse.software/faq/#custom-license), but in this case, I would suggest you make sure to clarify in the `README `file the context where this license is used and applied and that it refers to the binary component involved in booting Raspberry Pi devices. Something like "_The custom proprietary license in the LICENSES directory is part of the build system functionality and refers to closed source components involved in running or using a Raspberry Pi device"_

The copyright tag in the text file of this custom license itself refers only to the license text, so no other action is required.

Please bear in mind that this is meant to be a practical example of how the REUSE specifications would look like, feel free to implement them or some, but also feel free to proceed to do so on your own. 

I would like to note that while reaching REUSE compliance is a larger one-time chunk, maintaining this status is fairly simple: inclusion in CI pipelines, pre-commit hooks, badges, you-name-it, everything possible [^1].

Please also note that REUSE is an established practice with a lot of organisations using it, among them the KDE community, curl, GNUHealth (in progress), Linux kernel, companies such as Siemens, SAP, and LGE, as well as numerous smaller and larger projects. We would be happy to have you on board as well!

If there is something we can help with, feel free to reach out, we are more than happy to help!

[^0]: https://reuse.software/
[^1]: https://reuse.software/dev/

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,3 +1,10 @@
+<<<<<<< HEAD
+=======
+# SPDX-FileCopyrightText: meta-raspberrypi contributors
+#
+# SPDX-License-Identifier: MIT
+
+>>>>>>> 064cbc9 (changing license for two doc files)
 # Makefile for Sphinx documentation
 #
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,6 +1,13 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 #
+<<<<<<< HEAD
+=======
+# SPDX-FileCopyrightText: meta-raspberrypi contributors
+#
+# SPDX-License-Identifier: MIT
+#
+>>>>>>> 064cbc9 (changing license for two doc files)
 # meta-raspberrypi documentation build configuration file, created by
 # sphinx-quickstart on Tue May 23 09:51:24 2017.
 #
--- a/recipes-bsp/gpio-shutdown/files/gpio-shutdown-keymap.sh
+++ b/recipes-bsp/gpio-shutdown/files/gpio-shutdown-keymap.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
+
 ##
 # Bind the gpio-shutdown keycode as Keyboard signal and load it to the
 # keymap during startup.
--- a/recipes-graphics/vc-graphics/files/vchiq.sh
+++ b/recipes-graphics/vc-graphics/files/vchiq.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
+
 ### BEGIN INIT INFO
 # Provides:          vchiq.sh
 # Required-Start:    $remote_fs rmnologin
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: meta-raspberrypi contributors
+#
+# SPDX-License-Identifier: MIT
+
 build*
 *~
 .*.swp
--- a/LICENSES/CC-BY-4.0.txt
+++ b/LICENSES/CC-BY-4.0.txt
@@ -1,0 +1,156 @@
+Creative Commons Attribution 4.0 International
+
+ Creative Commons Corporation (“Creative Commons”) is not a law firm and does not provide legal services or legal advice. Distribution of Creative Commons public licenses does not create a lawyer-client or other relationship. Creative Commons makes its licenses and related information available on an “as-is” basis. Creative Commons gives no warranties regarding its licenses, any material licensed under their terms and conditions, or any related information. Creative Commons disclaims all liability for damages resulting from their use to the fullest extent possible.
+
+Using Creative Commons Public Licenses
+
+Creative Commons public licenses provide a standard set of terms and conditions that creators and other rights holders may use to share original works of authorship and other material subject to copyright and certain other rights specified in the public license below. The following considerations are for informational purposes only, are not exhaustive, and do not form part of our licenses.
+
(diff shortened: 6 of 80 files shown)


docs/Makefile
```diff
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: meta-raspberrypi contributors
+#
+# SPDX-License-Identifier: CC-BY-4.0
```
agherzan, line 3: Should this be MIT too?

docs/conf.py
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 #
+# SPDX-FileCopyrightText: meta-raspberrypi contributors
+#
+# SPDX-License-Identifier: CC-BY-4.0
```
agherzan, line 6: Should this be MIT too?

recipes-bsp/gpio-shutdown/files/gpio-shutdown-keymap.sh
```diff
@@ -1,4 +1,9 @@
 #!/bin/sh
+
+# SPDX-FileCopyrightText: meta-raspberrypi contributors
```
agherzan, line 3: Let's drop this header for now. I'd like to clarify its contribution first.

recipes-graphics/vc-graphics/files/vchiq.sh
```diff
@@ -1,4 +1,9 @@
 #!/bin/sh
+
+# SPDX-FileCopyrightText: meta-raspberrypi contributors
```
agherzan, line 3: Let's drop the header for this file for now - I'd like to check its contribution first.

agherzan (changes requested):
Hi @lnceballosz. First of all thanks for chasing this up. This is good progress and I'm generally happy with the recommendations here. Some comments on my side.

1. I would like to keep git history consistent so I'd reword `adding SPDX headers to doc files` -> `docs: Add SPDX headers`.
2. All your commits need to have a Signed-off-by. You can easily do that with the `-s` argument in `git commit` for example.
3. `creating LICENSES directory ` -> `Create LICENSES directory with the current licenses`
4. I would drop the dep5 file for now. My experience is that using it is always problematic as "umbrella" approaches are always error-prone. Let's leave the patches for now.
5. Rewording: `adding SPXD headers to .gitignore` -> `.gitignore: Add SPDX headers`. MInd the typo too.
6. Drop for now the header for recipes-graphics/vc-graphics/files/vchiq.sh - I need to figure it out first.
7. Squash all the other commits in one: "meta-raspberrypi: Add SPDX headers"

I don't think that the header of `vc-graphics.inc` and alike is a problem here because the meta-data can be MIT while the component itself is another LICENSE (even closed).

We already have a CI workflow that includes a reuse checker. See https://github.com/agherzan/meta-raspberrypi/blob/master/.github/workflows/compliance.yml#L37 . The only catch is that currently it is allowed to fail as the repository is not yet fully compliant. This MR will be a big step forward towards that. You will be able to see the CI logs once we cleanup this MR a bit and kick it.

recipes-bsp/gpio-shutdown/files/gpio-shutdown-keymap.sh
lnceballosz: Done

docs/Makefile
lnceballosz: also fixed

lnceballosz:
Hi @agherzan!
Uff what a journey, I have to say :) 
I'm not a git expert that's why I took a while to work on the changes you have requested. I think they are all addressed and solved. (Hope so) 
- I dropped the dep5 file, as you requested. FYI, we are currently working on changing this dep5 file for a yml file (still work in progress) which might make things way more clear, because I understand your position on this "umbrella" approach. So hopefully, we can bring this up in the coming months again with a new format :)
- I renamed the commits and signed them off. 
- And I have dropped those two headers that you requested and changed the license for those two doc files.
Hope this looks better, so if it does and of course up to you, feel free to merge this request. 
Once again, let me know if you have any questions or input now or in the near future. We will be very happy to keep supporting your project in becoming REUSE compliant. 

All the best,
Lina

agherzan:
This needs more work because the changes include git conflict marker that would break the layers.

PR #1096, opened 2022-10-03
rpi-cmdline: Leave cma value to kernel default

### LINKED ISSUE #1095: cma=64M in /boot/cmdline.txt causes libcamera-apps to fail for full resolution image capture
Is there any need to set `cma=64M` in `/boot/cmdline.txt` by default any more given that it [causes libcamera-apps to fail for full resolution image capture](https://github.com/raspberrypi/libcamera-apps/issues/375)?

https://github.com/agherzan/meta-raspberrypi/blob/c57b464b88769c1336695e48e1eb8f35e258b9ae/recipes-bsp/bootfiles/rpi-cmdline.bb#L16

tewarid:
Fixes #1095 where setting cma to a lower value affects the ability of libcamera-apps to capture higher resolution images.


agherzan (changes requested):
Just a small NIT on this - `cmdline.txt: leave cma value to kernel default ` -> `rpi-cmdline: Leave cma value to kernel default ` to be consistent with the current history format.

Issue #1117, opened 2022-11-30
VLC on RaspberryPI0w2 with framebuffer only


TomaszAIR:
**Description**

Want to use vlc without X11 and run it on fb only.

**Steps to reproduce the issue:**
1. Add `vlc` to IMAGE_INSTALL
2. Set machine to `raspberrypi0-2w`
3. Do not set x11 or wayland in DISTRO_FEATURES
4. Remove `vc4graphics` from MACHINE_FEATURES and use userland
5. Run build using kas-container v3.0.2 https://raw.githubusercontent.com/siemens/kas/3.0.2/kas-container

**Describe the results you received:**

We are using custom layer which uses meta-raspberrypi. Previously our project used omxplayer but we were told to try and switch to VLC as it gets more support https://github.com/agherzan/meta-raspberrypi/issues/1102.

rpidistro-vlc got some switches on x11 https://github.com/agherzan/meta-raspberrypi/blob/master/dynamic-layers/multimedia-layer/recipes-multimedia/rpidistro-vlc/rpidistro-vlc_3.0.12.bb#L53 so we try build it to run on fb. What we got are following build errors:

1. First is that running `bitbake ffmpeg` start ffmpeg from poky/meta

When force to build rpidistro-ffmpeg in return we get

```
ERROR: Nothing PROVIDES 'rpidistro-ffmpeg'
rpidistro-ffmpeg was skipped: incompatible with host arm-project-linux-gnueabi (not in COMPATIBLE_HOST)
```

This is probably because of https://github.com/agherzan/meta-raspberrypi/blob/master/recipes-multimedia/rpidistro-ffmpeg/rpidistro-ffmpeg_4.3.2.bb#L193, should not be there just `COMPATIBLE_HOST:rpi = "(.*)"` - so the same for https://github.com/agherzan/meta-raspberrypi/blob/master/dynamic-layers/multimedia-layer/recipes-multimedia/rpidistro-vlc/rpidistro-vlc_3.0.12.bb#L157 - after that `rpidistro-ffmpeg` starts build but fails on opengl check (mesa-gl successfully build with this fix https://github.com/agherzan/meta-raspberrypi/issues/987#issuecomment-1020306027)

```
| DEBUG: Python function extend_recipe_sysroot finished
| DEBUG: Executing shell function autotools_preconfigure
| DEBUG: Shell function autotools_preconfigure finished
| DEBUG: Executing python function autotools_aclocals
[...]

PR #1122, opened 2022-12-04
rpidistro-ffmpeg: fix opengl compile issue

EasyIP2023:
Connected to: https://github.com/agherzan/meta-raspberrypi/issues/1117

**Problem:**
rpidistro-ffmpeg fails the `./configure` stage if neither x11 or wayland defined
in `DISTRO_FEATURES`. After configure stage succeeds `compile` stage
fails as `SelectedGetProcAddress` isn't defined. One can't define it as
GLES2/gl2.h `glGetProcAddress` if not defined. Anywhere actually.

Also couldn't find the original locations of where I got the patches from, but if
you go to

https://github.com/RPi-Distro/ffmpeg/tree/ea72093f350f38edcd39c480b331c3219c377642/debian/patches

One can see that the patches applied were incorrect.

EDIT:

found the [ffmpeg-4.3.2-rpi_10.patch](https://github.com/RPi-Distro/ffmpeg/blob/378adea166f167a0e75cf85888f661101cb4af0f/debian/patches/ffmpeg-4.3.2-rpi_10.patch)

https://github.com/RPi-Distro/ffmpeg/tree/378adea166f167a0e75cf85888f661101cb4af0f

**Solution:**

Add two new patches so that `./configure` is OE ported and
`libavdevice/opengl_enc.c compiles` if `SelectedGetProcAddress` isn't
defined. It can't however be define as `glGetProcAddress` is not defined
anywhere. Patch applied loads GL functions pointers at compile time versus
dynamically at runtime.

Also MR was checked with all `./configure` flags. Only one to fail was gsm due to libgsm not found.

### DIFF
--- a/recipes-multimedia/rpidistro-ffmpeg/rpidistro-ffmpeg_4.3.2.bb
+++ b/recipes-multimedia/rpidistro-ffmpeg/rpidistro-ffmpeg_4.3.2.bb
@@ -33,23 +33,31 @@ RPROVIDES:${PN} = "${PROVIDES}"
 DEPENDS = "nasm-native"
 
 inherit autotools pkgconfig
-PACKAGECONFIG ??= "avdevice avfilter avcodec avformat swresample swscale postproc avresample \
-                   opengl udev sdl2 ffplay alsa bzlib lzma pic pthreads shared theora zlib \
-                   libvorbis x264 gpl sand rpi vout-drm vout-egl \
+PACKAGECONFIG ??= "avdevice avfilter avcodec avformat swresample swscale postproc avresample ffplay \
+                   sdl2 v4l2 drm udev alsa bzlib lzma pic pthreads shared theora zlib libvorbis x264 gpl \
+                   rpi sand vout-drm vout-egl \
                    ${@bb.utils.contains('MACHINE_FEATURES', 'vc4graphics', '', 'mmal', d)} \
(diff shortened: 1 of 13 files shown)


recipes-multimedia/rpidistro-ffmpeg/rpidistro-ffmpeg_4.3.2.bb
```diff
@@ -33,23 +33,31 @@ RPROVIDES:${PN} = "${PROVIDES}"
 DEPENDS = "nasm-native"
 
 inherit autotools pkgconfig
-PACKAGECONFIG ??= "avdevice avfilter avcodec avformat swresample swscale postproc avresample \
-                   opengl udev sdl2 ffplay alsa bzlib lzma pic pthreads shared theora zlib \
-                   libvorbis x264 gpl sand rpi vout-drm vout-egl \
+PACKAGECONFIG ??= "avdevice avfilter avcodec avformat swresample swscale postproc avresample ffplay \
```
agherzan, line 36: A fair bit of PACKAGECONFIG changes around. We will need to either split them in a separate commit or mention/explain them in git log.

EasyIP2023: Yes, closing this one. The next one should hopefully have a better explanation.

PR #1127, opened 2022-12-10
rpidistro-vlc: upgrade 3.0.12 -> 3.0.17

EasyIP2023:
Connected to: https://github.com/agherzan/meta-raspberrypi/issues/1117

Update is also meant to fix bug where if x11
isn't define in DISTRO_FEATURES builds will
fail to compile.

Upgrades version of VLC to 3.0.17
* Reason for update with version 3.0.12 compiling using GLESv2 and without x11/wayland defined in DISTRO_FEATURES it was proven to be difficult. Newer version of VLC appears to have better support for GLESv2 as there was far less hurdles.
* SRCREV set to b276eb0d7bc3213363e97dbb681ef7c927be6c73

Updates to PACKAGECONFIG
* Only include GLESv2 if opengl defined in DISTRO_FEATURES
* Only include libnotify is x11 included in DISTRO_FEATURES. Not checking for wayland as well because wayland integration requires a separate MR.
* remove ability to set certain flags as they require significant more development time to integrate. Probably best to include in separate MR. Flags for:
  - opencv
  - freerdp
  - samba
  - dvdread
  - vdpau
  - wayland
  - dca

Removal of

```
TARGET_CFLAGS:append = " -fcommon"
TARGET_CXXFLAGS:append = " -fcommon"
```

as define in configure.ac via the
3001-configure.ac-setup-for-OE-usage.patch patch

Removes extraneous sed operations in the do_configure takes. Appears to no longer be required.

Adds required flags to build that couldn't be set in configure.ac 

```
TARGET_CFLAGS:append = " -I${STAGING_INCDIR}/drm"
TARGET_LDFLAGS:append = " ${@bb.utils.contains('DISTRO_FEATURES',  'opengl', '-lGLESv2', '', d)}"
```

Replaces patches with updated patches used in actual commit. Prepend patches that are from meta-openembedded/meta-multimedia to 2XXX-<PATCH NAME>.patch. Adds nine new patches to fix ./configure, compile, runtime bugs.

PATCHES:
- 3001-configure.ac-setup-for-OE-usage.patch
* Used to ensure crossed compilted OE libraries are used during compilation. Specifically for the mmal and omxil libraries.

- 3002-fix-EGL-macro-undeclared-and-EGLImageKHR.patch
* Appears to be missing EGL preprocesor directives patch defines them. The function createImageKHR also appears to return EGLImageKHR not EGLImage patch updates. To accommodate for change.

- 3003-codec-omxil_core-replace-opt-vc-path-with-usr-lib.patch
* Configures omxil_core.c for OE usages as libbcm_host.so and libopenmaxil.so are located in a different location.

- 3004-use-GLESv2-headers-over-GL-headers.patch
* Ensures GLESv2 headers are used over OpenGL headers.

- 3005-modules-remove-glspectrum-usage.patch
* Removes glspectrum plugin as it appears to require normal OpenGL and not GLESv2 for usage.

- 3006-codec-omxil_core.h-fix-multiple-definition-of.patch
* Removes compile error due to function pointer in omxil_core.h header being defined multiple times as the header is included in multiple c source files.

- 3007-remove-xorg-related-link-libs.patch
* Removes xorg related flags if x11 isn't defined in DISTRO_FEATURES variable.

- 3008-vo-Makefile.am-exclude-libgl_plugin.patch
* Removes the libgl_plugin from being built if opengl not defined in DISTRO_FEATURES variable.

- 3009-vo-converter_vaapi-Fix-EGL-macro-undeclared.patch
* Appears to be missing EGL preprocesor directives patch defines them.


EasyIP2023:
@agherzan Sorry about the massive PR's Hopefully the commit message explains enough. Decided it would be best that way to save on unnecessary iterations.

Also, now it shouldn't matter what's set or not set in DISTRO_FEATURES variable. Should build just fine and display on screen, but the main distro features that I tested against where opengl x11 and wayland.

Have a branch for testing here https://github.com/EasyIP2023/meta-raspberrypi/tree/testing-two

agherzan (approved):
I think this is reasonable as it is pretty nicely documented. Added it up for CI and merge.

PR #741, opened 2020-10-06
Rpi eeprom upstream

macpijan:
Fixes: https://github.com/agherzan/meta-raspberrypi/issues/515

**- What I did**
- added rpi-eeprom recipe to allow EEPROM firmware update on the RPI4 devices
- added userlandtools recipe, which provides part of the userland tools required for EEPROM update

**- How I did it**
- inspired by the recipes from Balena
- more detailed description in the commit messages
- tested only on dunfell so far, that is why this PR targets dunfell

PR #1131, opened 2022-12-30
rpi-eeprom support

allanembedded:
Another attempt at merging rpi-eeprom so we can have the ability to update the rpi4 bootloader in yocto :)

This is mostly based on the previous recipe in pull requst #741, just updated to the latest upstream release which required a few dependency/licensing changes. It also uses the existing userland tools as I can't see a reason for it to use a different implementation that was in the previous pull request.

Two things to note are the unfortunate dependency on xxd which can pull in a lot of extra build time dependencies (enabling busybox xxd could be an alternative) and some shell errors printed during script execution which seem to be just differences in busybox execution, the update itself still seems to work fine.

As this is based on work previously done by @macpijan I've credited them with a Co-authored-by, happy to change that to something else if needed.


recipes-bsp/rpi-eeprom/rpi-eeprom_git.bb
```diff
@@ -0,0 +1,66 @@
...
+FILES:${PN} += "${base_libdir}/firmware/raspberrypi/bootloader/*"
+FILES:${PN} += "${sysconfdir}/*"
+
+INHIBIT_PACKAGE_STRIP = "1"
+INHIBIT_PACKAGE_DEBUG_SPLIT = "1"
+
+# vl805 tool sources are not available (yet), as it comes as a precompiled
+# binary only. It has ARM architecture whereas target machine is Aarch64. We
+# need to disable arch check for it otherwise it cannot packed.
+QAPATHTEST[arch] = ""
```
kraj, line 64: Have you tested this binary to work on rpi4 64 bit mode

allanembedded: Yes, it's a statically linked executable and it executes on aarch64 (assuming the kernel is configured to support 32-bit userspace applications).
```
% file tools/vl805
tools/vl805: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), statically linked, for GNU/Linux 3.2.0, BuildID[sha1]=1f534f3dce4bee014889300c7bb329ccda12b13c, stripped
```
```
root@raspberrypi4-64:~# uname -m    
aarch64
root@raspberrypi4-64:~# /usr/bin/vl805 -h
Usage: vl805 [<options>]
  where <options> are taken from:
    -h            Show this help message
    -r <file>     Read EEPROM
    -w <file>     Write EEPROM
    -v <file>     Verify EEPROM
    -o <offset>   Specify an offset to read from/write to (for testing).
                  N.B. Erases to the end of the EEPROM!
    -d            Verbose debugging
```

kraj: Statically linked ok makes sense

recipes-bsp/rpi-eeprom/rpi-eeprom_git.bb
```diff
@@ -0,0 +1,66 @@
...
+S = "${WORKDIR}/git"
+
+RDEPENDS:${PN} += " \
+    coreutils \
+    python3 \
+    python3-pycryptodomex \
+    openssl \
+    xxd \
+    pciutils \
+    userland \
```
agherzan, line 23: What's up with this dependency?

recipes-bsp/rpi-eeprom/rpi-eeprom_git.bb
```diff
@@ -0,0 +1,66 @@
...
+    ln -s stable ${D}${base_libdir}/firmware/raspberrypi/bootloader/latest
+
+    # copy default config
+    install -d ${D}${sysconfdir}/default
+    install -D ${S}/rpi-eeprom-update-default ${D}${sysconfdir}/default/rpi-eeprom-update
+}
+
+FILES:${PN} += "${sbindir}/vl805"
+FILES:${PN} += "${base_libdir}/firmware/raspberrypi/bootloader/*"
+FILES:${PN} += "${sysconfdir}/*"
```
agherzan, line 56: Isn't this in by default? https://git.yoctoproject.org/poky/tree/meta/conf/bitbake.conf#n327

recipes-bsp/rpi-eeprom/rpi-eeprom_git.bb
```diff
@@ -0,0 +1,66 @@
...
+
+    ln -s critical ${D}${base_libdir}/firmware/raspberrypi/bootloader/default
+    ln -s stable ${D}${base_libdir}/firmware/raspberrypi/bootloader/latest
+
+    # copy default config
+    install -d ${D}${sysconfdir}/default
+    install -D ${S}/rpi-eeprom-update-default ${D}${sysconfdir}/default/rpi-eeprom-update
+}
+
+FILES:${PN} += "${sbindir}/vl805"
```
agherzan, line 54: I would expect this to be in by default. Isn't that the case?

agherzan (changes requested):
Left some comments. A tiny NIT too: ` rpi-eeprom: Add recipe for eeprom update/configuration` -> ` rpi-eeprom: Add recipe for eeprom update/configuration`

recipes-bsp/rpi-eeprom/rpi-eeprom_git.bb
allanembedded: Assuming you mean userland, the userland package provides the vcgencmd command which rpi-eeprom-config and rpi-eeprom-update can use to query the video core for the current bootloader configuration and version. I say "can" because the scripts do appear to support alternative methods of finding this information, but when executing the scripts they end up falling back to vcgencmd.

For a single example, see this hunk of code from rpi-eeprom-config.
```python
def read_current_config():
    """
    Reads the configuration used by the current bootloader.
    """
    fw_base = "/sys/firmware/devicetree/base/"
    nvmem_base = "/sys/bus/nvmem/devices/"

    if os.path.exists(fw_base + "/aliases/blconfig"):
        with open(fw_base + "/aliases/blconfig", "rb") as f:
            nvmem_ofnode_path = fw_base + f.read().decode('utf-8')
            for d in os.listdir(nvmem_base):
                if os.path.realpath(nvmem_base + d + "/of_node") in os.path.normpath(nvmem_ofnode_path):
                    return (open(nvmem_base + d + "/nvmem", "rb").read().decode('utf-8'), "blconfig device")

    return (shell_cmd(['vcgencmd', 'bootloader_config']), "vcgencmd bootloader_config")
```

In a default build using langdale, the nodes referenced in /sys/firmware/devicetree/base/ do exist, however /sys/bus/nvmem/devices/ is empty. For clarity, I've not done any serious investigation here since vcgencmd enables the scripts to work as-is.

recipes-bsp/rpi-eeprom/rpi-eeprom_git.bb
allanembedded: Removed in latest revision.

recipes-bsp/rpi-eeprom/rpi-eeprom_git.bb
allanembedded: Removed in latest revision.

agherzan (changes requested):
Blocking it for now until we clarify the `userland`  dependency.

recipes-bsp/rpi-eeprom/rpi-eeprom_git.bb
agherzan: (besides this, everything else looks fine to me now)

agherzan: Your fix was merged. So that looks good. For the kernel config, I'm happy to start pulling in CONFIG_NVMEM_RMEM unconditionally for the relevant machines. So I'm up for a new cfg.

allanembedded: OK, I've respun the branch with the following changes:

* Rebased branch onto latest master (tested with mickledore due to master changes).
* Updated rpi-eeprom busybox patch to say that it's a backport (merged but not yet part of a tagged release).
* Added config fragment for built in nvmem-rmem on raspberrypi4 machines.
* Updated rpi-eeprom commit message to explain the nvmem-rmem dependency.
* Tested on 5.15 kernel for raspberrypi4 and raspberrypi4-64.

agherzan (approved):
This looks nice and clean to me now. Happy to start the CI rig on it.

recipes-bsp/rpi-eeprom/rpi-eeprom_git.bb
agherzan: Looks good.

djr-spectrummfg:
Can this be merged to kirkstone? Since kirkstone is the current Yocto LTS branch.

agherzan:
@djr-spectrummfg Absolutely. I'm happy to accept a merge request that cherry-picks these exact commits into Kirkstone.

PR #1134, opened 2023-01-05
linux-raspberrypi.inc: Weakly assign COMPATIBLE_MACHINE

kraj:
COMPATIBLE_MACHINE is expected to be overridden by linux-raspberrypi-v7.inc for linux-raspberrypi-v7 machine this however does not work because linux-raspberrypi-v7_5.15.bb includes linux-raspberrypi-v7.inc and linux-raspberrypi.inc ( via linux-raspberrypi_5.15.bb) and setting from linux-raspberrypi.inc takes precedence over one coming from linux-raspberrypi-v7.inc as a result linux-raspberrypi-v7 kernel recipe becomes available for raspberrypi4 machine as well, this recipe however needs certain defines e.g.RASPBERRYPI_v7_KERNEL which are only available in raspberrypi-armv7.conf


agherzan (approved):
Nice. Approved and marked for merge.

PR #1149, opened 2023-02-12
linux-firmware-rpidistro: Create brcmfmac43430-sdio.raspberrypi,model…

fscr:
…-zero-w.bin symlink

This fixes a boot time message on rpi0w
brcmfmac mmc1:0001:1: Direct firmware load for brcm/brcmfmac43430-sdio.raspberrypi,model-zero-w.bin failed with error -2


agherzan (changes requested):
Thanks for this. Would you be able to change the git commit in accordance with https://meta-raspberrypi.readthedocs.io/en/latest/contributing.html? Also, git history on that specific recipe should give you good examples.

fscr:
> Thanks for this. Would you be able to change the git commit in accordance with https://meta-raspberrypi.readthedocs.io/en/latest/contributing.html? Also, git history on that specific recipe should give you good examples.

Actually I checked that page before and the history and then adapted the commit message of commit a5c2bef10eb22acd8a1e3c2647c7df41e0931f4d which added the same link for model 4b.

Now I added also a similar link for 3B which probably faces the same issue and adapted the commit message.

Another idea about this recipe would be to create the symlinks only for the target MACHINE, resp. all in the set raspberrypi-armv7 or raspberrypi-armv8. For me the current change is sufficient though.

Do I have to trigger the approval workflow somehow or is it done by maintainers only?

agherzan (approved):

agherzan:
@fscr Some NITS for net times ( and it will help in other projects too)
1. keep line length reasonable including subject
2. subject doesn't end in full stop

Thanks for this.

PR #1152, opened 2023-02-27
python3-adafruit-blinka: Fix the correct python recipes path in dynam…

mdrodrigo:
…ic-layer sub-dir

The python recipes in the dynamic-layer need one more subdirectory to be in accordance with the defined in the layer settings.

The recipes moved to new python sub-dir are:
python3-adafruit-blinka
python3-adafruit-circuitpython-busdevice
python3-adafruit-circuitpython-motor
python3-adafruit-circuitpython-motorkit
python3-adafruit-circuitpython-pca9685

**- What I did**

**- How I did it**


agherzan (approved):
Thanks. This looks good.

PR #1154, opened 2023-03-02
Enable build of mesa-demos for userland graphics drivers

clopez:
Enable build of mesa-demos for userland graphics drivers
* mesa-demos can be built fine if both `gles1` and `egl` options are disabled. This allows to have utilities like `glxinfo` or `glxgears`


clopez:
,mm, this is going to fail to build if the `wayland` package option gets enabled (and it gets enabled if the distro has the feature) unless poky contains this patch that [I just sent to the oe-core mailing list](https://lists.openembedded.org/g/openembedded-core/message/177973)

I guess is better that we disable wayland here as well. On this package enabling packageconfig option  `wayland` without `egl` is useless in the end.

agherzan (changes requested):
Looks good. Thanks for this. The only request would be to keep the current git log format:

https://meta-raspberrypi.readthedocs.io/en/latest/contributing.html

clopez:
> Looks good. Thanks for this. The only request would be to keep the current git log format:
> 
> https://meta-raspberrypi.readthedocs.io/en/latest/contributing.html

Pushed a new commit with that format.

Thanks!

agherzan (approved):

PR #1172, opened 2023-04-29
Add support for Raspberry Pi Camera Module v3 to Kirkstone

Jessseee:
@floion is requesting at [meta-balena](https://github.com/balena-os/balena-raspberrypi/issues/999#issuecomment-1528739563) for the changes of #1158 to be applied to meta-raspberrypi/kirkstone as well.


agherzan (approved):
Looks good.

Jessseee:
It seems the build has run into the issue mentioned by @quaresmajose on #1158.

quaresmajose:
The most reasonable thing to do IMO is backporting the [imx708](https://github.com/raspberrypi/linux/pull/5452) on the kernel 5.10 but please note that it build but doesn't work on this kernel so maybe a note will be needed about that.
In this way kernel 5.10 still be usable and the camera could be used on the 5.15.

Jessseee:
Is it necessary for both kernel version 5.10 and 5.15 to be supported on the Kirkstone branch, as 5.10 is also on Holister, Hardknott and Gatesgarth? It appears that the differences between 5.10 and 5.15 seem to be significant enough that maybe they should not be supported at the same time.

Jessseee:
@agherzan what are your thoughts on this?

agherzan:
@quaresmajose That would be just masking an issue. So I'm not sure if I'm on board with that.
@Jessseee The only clean way I see is to keep the change in master and backport the change on kirkstone in a separate layer. And master to drop 5.10.

Jessseee:
I will try to make a separate layer. Closing this PR for now.

PR #1181, opened 2023-05-25
99-com.rules: Update rules for BCM2711 based hardware

floion:
Update the rules as per
https://github.com/RPi-Distro/raspberrypi-sys-mods/commit/05cfe136f76677a891a0c77dc91d796913607c07

**- What I did**

**- How I did it**

### DIFF
--- a/recipes-core/udev/udev-rules-rpi/99-com.rules
+++ b/recipes-core/udev/udev-rules-rpi/99-com.rules
@@ -1,8 +1,21 @@
-KERNEL=="ttyAMA[01]", PROGRAM="/bin/sh -c '\
+KERNEL=="ttyAMA0", PROGRAM="/bin/sh -c '\
 	ALIASES=/proc/device-tree/aliases; \
 	if cmp -s $$ALIASES/uart0 $$ALIASES/serial0; then \
 		echo 0;\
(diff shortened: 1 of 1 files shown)


floion:
So you mean adding 99-com.rules from https://github.com/agherzan/meta-raspberrypi/blob/master/recipes-core/udev/udev-rules-rpi/99-com.rules in https://github.com/agherzan/meta-raspberrypi/blob/master/recipes-core/udev/udev-rules-rpi.bb as SRC_URI and pull it from there?
What about https://github.com/agherzan/meta-raspberrypi/blob/master/recipes-core/udev/udev-rules-rpi/can.rules ? Would this be installed by the same `udev-rules-rpi` recipe?

floion:
@agherzan you were thinking something along the lines of https://github.com/agherzan/meta-raspberrypi/pull/1185 ?

agherzan:
Exactly. That looks better to me. And maybe in the future split that in its recipe with some deps (to maintain compatibility). So that the rpi udev recipe brings only custom ones and pulls in the raspberry-pi-sys-mods. But for now, this is a good step to avoid any surprises.

floion:
Superseded in favor of https://github.com/agherzan/meta-raspberrypi/pull/1185

PR #1185, opened 2023-05-26
udev-rules-rpi: Use 99-com.rules directly from upstream

floion:
Rather than have an unsynchronized copy of this file, let's use this udev rule directly from upstream.

**- What I did**

**- How I did it**


agherzan (approved):
Thanks, @floion!

floion:
@agherzan no worries
Anything else needed to get this merged?

Issue #1189, opened 2023-05-31
RPi4 does not boot when enabling RPI_USE_U_BOOT


5o5uk3:
**Description**

When enabling `RPI_USE_U_BOOT` in `conf/local.conf`, the raspberry pi 4 does not boot *(`langdale` branch)*.

**Steps to reproduce the issue:**
1. Clone the yocto project
2. Add the meta-raspberrypi layer
3. Add `RPI_USE_U_BOOT = "1"` in `conf/local.conf` file
4. build rpi-test-image and copy it to a SD card
5. boot

**Current results:**

* Nothing on HDMI output / only red led is on - green led remains off
* Nothing on UART but that's normal as it is disabled

**Expected results:**

* A normal boot sequence

**Howto fix:**

* A simple workaround is to add `enable_uart=1` in `config.txt` file. The reason is explained in `arch/arm/mach-bcm283x/Kconfig` from u-boot source code:

```
 config TARGET_RPI_4
    bool "Raspberry Pi 4 64-bit build"
    help
      Support for all BCM2711-based Raspberry Pi variants, such as
      the RPi 4 model B, in AArch64 (64-bit) mode.

      This option assumes the VideoCore firmware is configured to use the
      mini UART (rather than PL011) for the serial console. This is the
      default on the RPi 4. To enable the UART console, the following non-
      default option must be present in config.txt: enable_uart=1. This is
      required for U-Boot to operate correctly, even if you only care
      about the HDMI/usbkbd console.
```

Note that this requirement is also valid for other raspberry pi flavors.
IMHO, `recipes-bsp/bootfiles/rpi-config_git.bb` should add a conditional statement like:

```bash
    elif [ "${RPI_USE_U_BOOT}" = "1" ] ; then
        echo "# Enable UART" >>$CONFIG
        echo "enable_uart=1" >>$CONFIG
```

Indeed, it would be a bad idea to set `ENABLE_UART = "1"` when `RPI_USE_U_BOOT` is enabled as it would also add the serial console parameter on the kernel command line which we may not want.

agherzan:
That is an excellent catch. I was not aware of this. Your change is reasonable for me. You mention that this is required for all boards. Why is that, exactly? The snippet above only mentions RPi4-64.

5o5uk3:
[...]

PR #1190, opened 2023-05-31
U-Boot requires "enable_uart=1" for various boards to operate correctly.

5o5uk3:
** Summary **

* When RPI_USE_U_BOOT is set to 1 in conf/local.conf, if enable_uart is not set to 1 in config.txt then the board won't boot (Resolves agherzan/meta-raspberrypi#1189).

**- What I did**

* Add a new section to check the requirement related to `enable_uart=1` in the recipe which generates the system parameters aka `config.txt` file.
* I've added a new section to keep the code readable because I also check if `enable_uart` has been set through `ENABLE_UART` environment variable. Indeed, I tried to avoid duplicating lines in `config.txt` and above all I tried to detect opposing configuration i.e. `ENABLE_UART = "0"` vs `enable_uart=1`.

**-What is missing**

The specific raspberry pi 2 case and its specific `dtoverlay=pi3-miniuart-bt` configuration is not handled.

**-Tests performed**

* I've rebuilt my image for RPi4-64 with the following configurations:
  * RPI_USE_U_BOOT = "1" / ENABLE_UART = "1" => OK *(single line `enable_uart=1` in config.txt)*
  * RPI_USE_U_BOOT = "1" / ENABLE_UART not set => OK
  * RPI_USE_U_BOOT = "1" / ENABLE_UART = "0" => OK *(got a build failure => `do_deploy: Invalid configuration: RPI_USE_U_BOOT requires to enable UART in config.txt for raspberrypi-armv8`)*
  * RPI_USE_U_BOOT = "0" / ENABLE_UART = "1" => OK
  * RPI_USE_U_BOOT = "1" / ENABLE_UART not set => OK *(no `enable_uart` in config.txt)*
  *  RPI_USE_U_BOOT = "0" / ENABLE_UART = "0" => OK


recipes-bsp/bootfiles/rpi-config_git.bb
```diff
@@ -178,13 +178,26 @@ do_deploy() {
...
     # UART support
-    if [ "${ENABLE_UART}" = "1" ] || [ "${ENABLE_UART}" = "0" ] ; then
+    if [ "${ENABLE_UART}" = "1" ] || [ "${ENABLE_UART}" = "0" ]; then
         echo "# Enable UART" >>$CONFIG
         echo "enable_uart=${ENABLE_UART}" >>$CONFIG
     elif [ -n "${ENABLE_UART}" ]; then
         bbfatal "Invalid value for ENABLE_UART [${ENABLE_UART}]. The value for ENABLE_UART can be 0 or 1."
     fi
 
+    # U-Boot requires "enable_uart=1" for various boards to operate correctly
```
agherzan, line 188: This looks a bit redundant. Can we simplify it by just bbfatal (without any config changes) when we detect a misconfiguration? In this way we flag and users can fix in their conf.

```
if [ "${RPI_USE_U_BOOT}" = "1" ] && [ "${ENABLE_UART}" != "1" ]; then
  case "${UBOOT_MACHINE}" in
    rpi_0_w_defconfig|rpi_3_32b_config|rpi_4_32b_config|rpi_arm64_config)
      bbfatal "Invalid configuration: RPI_USE_U_BOOT requires to enable the UART in config.txt for ${MACHINE}"
      ;;
  esac
fi
```

agherzan (changes requested):
I would also mind the commit format https://meta-raspberrypi.readthedocs.io/en/latest/contributing.html

recipes-bsp/bootfiles/rpi-config_git.bb
agherzan: I see. This makes sense now on a second review. I'm okay with it now. Could you tweak git log a bit, as per the other comment?

agherzan, line 188: Could you also add here a link reference to he config in uboot tree?

agherzan:
Looks good. I've approved and marked for automerge.

recipes-bsp/bootfiles/rpi-config_git.bb
```diff
@@ -178,13 +178,27 @@ do_deploy() {
...
+    # cf https://source.denx.de/u-boot/u-boot/-/blob/v2023.04/arch/arm/mach-bcm283x/Kconfig?ref_type=tags#L65
+    if [ "${RPI_USE_U_BOOT}" = "1" ] && [ "${ENABLE_UART}" != "1" ]; then
+        case "${UBOOT_MACHINE}" in
+            rpi_0_w_defconfig|rpi_3_32b_config|rpi_4_32b_config|rpi_arm64_config)
+                if [ "${ENABLE_UART}" = "0" ]; then
+                    bbfatal "Invalid configuration: RPI_USE_U_BOOT requires to enable the UART in config.txt for ${MACHINE}"
```
floion, line 194: @agherzan it seems that because of this bbfatal the following lines of code never get executed. So "enable_uart" does not get to be set to 1 in config.txt

The above is one thing. The second one is that we actually did not have any issues like those described here

PR #1194, opened 2023-06-21
*.patch: add Upstream-Status to all patches

shr-project:
There is new patch-status QA check in oe-core:
https://git.openembedded.org/openembedded-core/commit/?id=76a685bfcf927593eac67157762a53259089ea8a

This is temporary work around just to hide _many_ warnings from optional patch-status (if you add it to WARN_QA).

This just added
Upstream-Status: Pending
everywhere without actually investigating what's the proper status.

This is just to hide current QA warnings and to catch new .patch files being added without Upstream-Status, but the number of Pending patches is now terrible: Patches in Pending state: 41 (57%)

With recent change to enable patch-status not only for all .patch files in oe-core, but for all recipes from oe-core: https://git.openembedded.org/openembedded-core/commit/?id=61a881fdbe8b5a21c6276b8a5d06cc30486b1eb3 this causes bluez5 do_patch failures as reported in: https://lists.openembedded.org/g/openembedded-core/message/183177

**- What I did**

**- How I did it**


agherzan (approved):
Agreed. Thanks for the patch.

PR #1197, opened 2023-06-29
Add support for Raspberry Pi Camera Module v3 to Kirkstone (compatibility patch)

Jessseee:
The imx708 devicetree layer is required for use of the Raspberry Pi Camera Module v3. This was implemented in #1158 on master. However, the dts file is not included with Kernel 5.10. Therefore, to port these changes to Kirkstone, I added a dynamic layer to install the imx708 devicetree overlay when building for Linux kernel 5.10. Also see #1172 for more context.

### DIFF
--- a/recipes-kernel/linux/files/0003-recipe-bsp-Add-imx708-dt-overlay-files.patch
+++ b/recipes-kernel/linux/files/0003-recipe-bsp-Add-imx708-dt-overlay-files.patch
@@ -1,0 +1,196 @@
+From a66bb44e8c32170d2de5ea29ae16b30d9a5e2b06 Mon Sep 17 00:00:00 2001
+From: Jesse Visser <jssvssr2000@gmail.com>
+Date: Thu, 9 Mar 2023 11:19:34 +0100
+Subject: [PATCH] recipe-bsp: Add imx708 dt-overlay files
+
+Signed-off-by: Jesse Visser <jssvssr2000@gmail.com>
+---
+ recipes-kernel/linux/files/imx708-overlay.dts | 105 ++++++++++++++++++
+ recipes-kernel/linux/files/imx708.dtsi        |  35 ++++++
+ recipes-kernel/linux/linux-raspberrypi-dev.bb |   2 +
+ .../linux/linux-raspberrypi_5.10.bb           |   2 +
+ 4 files changed, 144 insertions(+)
+ create mode 100644 recipes-kernel/linux/files/imx708-overlay.dts
+ create mode 100644 recipes-kernel/linux/files/imx708.dtsi
+
+diff --git a/recipes-kernel/linux/files/imx708-overlay.dts b/recipes-kernel/linux/files/imx708-overlay.dts
+new file mode 100644
+index 0000000..6efbe09
+--- /dev/null
++++ b/recipes-kernel/linux/files/imx708-overlay.dts
+@@ -0,0 +1,105 @@
++// SPDX-License-Identifier: GPL-2.0-only
++// Definitions for IMX708 camera module on VC I2C bus
++/dts-v1/;
++/plugin/;
++
++#include <dt-bindings/gpio/gpio.h>
++
++/{
++	compatible = "brcm,bcm2835";
++
++	fragment@0 {
++		target = <&i2c0if>;
++		__overlay__ {
++			status = "okay";
++		};
++	};
++
++	clk_frag: fragment@1 {
++		target = <&cam1_clk>;
++		__overlay__ {
++			status = "okay";
++			clock-frequency = <24000000>;
++		};
++	};
++
++	fragment@2 {
++		target = <&i2c0mux>;
++		__overlay__ {
++			status = "okay";
++		};
++	};
++
++	reg_frag: fragment@3 {
++		target = <&cam1_reg>;
++		cam_reg: __overlay__ {
++			startup-delay-us = <70000>;
++			off-on-delay-us = <30000>;
++			regulator-min-microvolt = <2700000>;
++			regulator-max-microvolt = <2700000>;
++		};
++	};
++
++	fragment@4 {
++		target = <&cam_node>;
++		__overlay__ {
++			lens-focus = <&vcm_node>;
++		};
++	};
++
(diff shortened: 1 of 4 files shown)


agherzan:
Could you add the rationale and context in the description?

agherzan:
Reopened as I've closed it by mistake.

Jessseee:
> Could you add the rationale and context in the description?

I hope this is clear enough.

mhazley:
@Jessseee made it here from the Balena thread? Any way we can help get this over the line, we'd like to use camera v3 on Balena OS on  Pi4

agherzan:
I'm generally happy with this, but we will need to first land it via a review in master. It looks good to me otherwise.

dynamic-layers/imx708-layer/recipes-kernel/linux/linux-raspberrypi-dev.bbappend
```diff
@@ -0,0 +1,4 @@
+SRC_URI += "file://imx708-overlay.dts;subdir=git/arch/${ARCH}/boot/dts/overlays"
+
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
```
agherzan, line 4: I'm not sure what this is all about in the imx708 layer. Shouldn't this be a bbappend in our kernel recipes? Also, isn't this upstreamable?

agherzan (changes requested):

agherzan:
That's what I hoped and thought, but this seems to include more things. @Jessseee can you explain the backport so I understand where we are with this?

mhazley:
> but this seems to include more things

Sorry, I completely missed that!

Jessseee:
I am quite new to working with yocto, so I am not quite sure what the best approach is. However, my idea was that because the dts file is completely missing from Kernel 5.10 it has to be externally added.

agherzan:
That makes sense @Jessseee. I think the best would be to introduce it in the kernel recipes we have in the layer - linux-raspberrypi ones.

mhazley:
@Jessseee I _think_ you can just make your bbappend for the dts `recipes-kernel/linux/linux-raspberrypi_%.bbappend` and it will apply the append to both the 5.10 and the 5.15 recipes.

I am quite new to yocto too, but I patched a dtoverlay in a similar way in BalenaOS once.

Jessseee:
@mhazley The problem is that the files should only be added to `linux-raspberrypi-dev` and `raspberrypi-5.10`, because I think you will run into issues with 5.15 as it already has the driver. As I think @agherzan said, they need to be added directly to the kernel recipes.

ScottGibb:
@agherzan Is there any update on this approved? Also thank you for doing the work and adding it to the meta-raspberrypi layer!

agherzan:
@Jessseee This looks good. One last request: can you please squash "imx708: Add compatibility layer " into "recipe-bsp: Add support for Raspberry Pi Camera Module v3". They make sense as one commit. Otherwise, I think we are good to go.

Jessseee:
I squashed the commits and updated the branch.

recipes-kernel/linux/linux-raspberrypi-dev.bb
```diff
@@ -24,6 +24,8 @@ SRC_URI = " \
     git://git.yoctoproject.org/yocto-kernel-cache;type=kmeta;name=meta;branch=${LINUX_RPI_KMETA_BRANCH};destsuffix=${KMETA} \
     file://powersave.cfg \
     file://android-drivers.cfg \
+    file://imx708-overlay.dts;subdir=git/arch/${ARCH}/boot/dts/overlays \
```
agherzan, line 27: Interesting approach. @kraj does this have any downsides? I don't see any - maybe when using devtool?

kraj: I think it might be better to have it as a patch, if it gets added in one of future kernels then we wont be overwriting it without knowing.

agherzan: That is a fair point. @Jessseee do you reckon you can update this with a patch?

Jessseee: Yes, I changed the addition of the imx708 dt-overlay to a patch.

recipes-kernel/linux/files/0003-recipe-bsp-Add-imx708-dt-overlay-files.patch
```diff
@@ -0,0 +1,196 @@
...
++	};
++};
++
++vcm_node: dw9817@c {
++	compatible = "dongwoon,dw9817-vcm";
++	reg = <0x0c>;
++	status = "disabled";
++	VDD-supply = <&cam1_reg>;
++};
+diff --git a/recipes-kernel/linux/linux-raspberrypi-dev.bb b/recipes-kernel/linux/linux-raspberrypi-dev.bb
```
agherzan, line 168: I don't think this was supposed to be here.

recipes-kernel/linux/files/0003-recipe-bsp-Add-imx708-dt-overlay-files.patch
```diff
@@ -0,0 +1,196 @@
+From a66bb44e8c32170d2de5ea29ae16b30d9a5e2b06 Mon Sep 17 00:00:00 2001
```
agherzan, line 1: I might have been misunderstood. I was proposing to have this as a commit over the kernel source. Ideally, a formatted patch directly from rpi Linux kernel upstream. Does that make sense? Also, mind Upstream-status footer.

agherzan: To be clear, I expect:
1. Cherry-pick https://github.com/agherzan/meta-raspberrypi/pull/1158
2.  Add a commit that adds a kernel patch to include the respective dtb (ideally as clean/backported patch from upstream)

agherzan (changes requested):
Thanks for chasing this up.

agherzan:

PR #1199, opened 2023-07-22
kas-poky-rpi.yml: renamed ABORT to HALT

vermaete:
Naming changed in Yocto.  And it avoids a warning during the run.


agherzan (approved):
Looks good. Thanks.

PR #1201, opened 2023-07-25
rpi-base.inc: add the disable-wifi device tree overlay

vermaete:
It can be useful to disable the WiFi for hardening reasons. And it probably lower the power consumption a bit.

### DIFF
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -18,6 +18,7 @@ RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
     overlays/overlay_map.dtb \
     overlays/at86rf233.dtbo \
     overlays/disable-bt.dtbo \
+    overlays/disable-wifi.dtbo \
     overlays/dwc2.dtbo \
     overlays/gpio-ir.dtbo \
     overlays/gpio-ir-tx.dtbo \


conf/machine/include/rpi-base.inc
```diff
@@ -18,6 +18,7 @@ RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
     overlays/overlay_map.dtb \
     overlays/at86rf233.dtbo \
     overlays/disable-bt.dtbo \
+    overlays/disable-wifi.dtbo \
```
kraj, line 21: does it change just the out of box defaults ?

vermaete: No, the image before and after the change is the same.  Except that the dtbo file is in the /boot partition.  But not used.

If you would like to have the Wifi disabled you should add into the kas file:
`RPI_EXTRA_CONFIG = "dtoverlay=disable-wifi"`

If you would like to have Bluetooth and Wifi disabled you can use
`RPI_EXTRA_CONFIG = "dtoverlay=disable-wifi\ndtoverlay=disable-bt"`

kraj: yeah thanks for confirming.

kraj (approved):
lgtm

agherzan (approved):

PR #1202, opened 2023-07-25
rpi-libcamera-apps: fix Illegal Instruction

vivien:
With armv8-neon being enforced for all ARM based machines, running any libcamera app on raspberrypi (e.g. Raspberry Pi B Rev 2) results in Illegal Instruction.

To fix this, defaults to TARGET_ARCH for non-aarch64 machines.


kraj (approved):
lgtm

agherzan (approved):
Thanks!

agherzan:
Isn't this affecting master/main?

agherzan:
@vivien can you please submit it for master and afterwards cherry pick + MR for mick?

vivien:
> @vivien can you please submit it for master and afterwards cherry pick + MR for mick?

@agherzan I would argue that is the responsibility of the maintainer, not the developer. Also there is a better approach to this fix actually, so I will send a PR against master along with additional changes.

kraj:
> > @vivien can you please submit it for master and afterwards cherry pick + MR for mick?
> 
> @agherzan I would argue that is the responsibility of the maintainer, not the developer. Also there is a better approach to this fix actually, so I will send a PR against master along with additional changes.

Its perhaps everyone's interest to help maintainer's load as much as possible to keep things in good shape and especially when they ask for it. Generally projects workflows route stable fixes via trunk so even if something is observed on a release branch the process is to
validate it on trunk and in some cases it may not be an issue for trunk but that is to be documented in release branch process.

agherzan:
I would enforce what @kraj wrote above. We can discuss that cherry-picking on stable branches is part of maintainership (even though we are a community, so contributions are always helpful). Still, the changes/fixes are generally sent to the main branch first. And that is part of the public contribution flow. Once that lands, it is in the contributor's interest to test and propose a backport. I hope that makes sense, and thank you for your work.

vivien:
If that part of the maintenance has to be handled by the developers as well, that's fine with me, I'll be happy to do that.

agherzan:
Thanks, @vivien.

PR #1177, opened 2023-05-19
Enable dtoverlay=imx477 when RASPBERRYPI_HD_CAMERA is set

leigh-johnson:
**- What I did**

Correctly handle `RASPBERRYPI_HD_CAMERA` when writing out `/boot/config.txt`

Looks like this code block was commented out, but should be enabled. =)

Closes: https://github.com/agherzan/meta-raspberrypi/issues/1167
Ref: https://github.com/bitsy-ai/printnanny-os/issues/311

PR #1213, opened 2023-08-15
Add officially supported camera modules

vivien:
This patchset adds support for the camera modules listed in the official [documentation](https://www.raspberrypi.com/documentation/computers/camera_software.html#if-you-do-need-to-alter-the-configuration) and simplify the configuration with a single RASPBERRYPI_CAMERA variable (and predefined aliases) to describe if and which camera model is used.

At the same time, add an empty config.txt template file to the rpi-config package, because relying on a commented out config file from a user github repository is error prone and only a few keys are added to the configuration anyway. Prefer to point the user to the official [documentation](https://www.raspberrypi.com/documentation/computers/config_txt.html) instead.

### DIFF
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -31,6 +31,10 @@ RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
     overlays/i2c-gpio.dtbo \
     overlays/i2c-rtc.dtbo \
     overlays/imx219.dtbo \
+    overlays/imx290.dtbo \
+    overlays/imx296.dtbo \
+    overlays/imx327.dtbo \
+    overlays/imx378.dtbo \
     overlays/imx477.dtbo \
     overlays/imx708.dtbo \
     overlays/iqaudio-dac.dtbo \
@@ -42,6 +46,8 @@ RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
     overlays/mcp2515-can1.dtbo \
     overlays/mcp3008.dtbo \
     overlays/miniuart-bt.dtbo \
+    overlays/ov5647.dtbo \
+    overlays/ov9281.dtbo \
     overlays/overlay_map.dtb \
     overlays/pitft22.dtbo \
     overlays/pitft28-capacitive.dtbo \
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -15,7 +15,6 @@ XSERVER = " \
     "
 
 RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
-    overlays/overlay_map.dtb \
     overlays/at86rf233.dtbo \
     overlays/disable-bt.dtbo \
     overlays/disable-wifi.dtbo \
@@ -29,9 +28,6 @@ RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
     overlays/hifiberry-dac.dtbo \
     overlays/hifiberry-dacplus.dtbo \
     overlays/hifiberry-digi.dtbo \
-    overlays/justboom-both.dtbo \
-    overlays/justboom-dac.dtbo \
-    overlays/justboom-digi.dtbo \
     overlays/i2c-gpio.dtbo \
     overlays/i2c-rtc.dtbo \
     overlays/imx219.dtbo \
@@ -39,10 +35,14 @@ RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
     overlays/imx708.dtbo \
     overlays/iqaudio-dac.dtbo \
     overlays/iqaudio-dacplus.dtbo \
+    overlays/justboom-both.dtbo \
+    overlays/justboom-dac.dtbo \
+    overlays/justboom-digi.dtbo \
     overlays/mcp2515-can0.dtbo \
     overlays/mcp2515-can1.dtbo \
     overlays/mcp3008.dtbo \
     overlays/miniuart-bt.dtbo \
+    overlays/overlay_map.dtb \
     overlays/pitft22.dtbo \
     overlays/pitft28-capacitive.dtbo \
     overlays/pitft28-resistive.dtbo \
@@ -52,9 +52,9 @@ RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
     overlays/rpi-poe.dtbo \
     overlays/vc4-fkms-v3d.dtbo \
     overlays/vc4-fkms-v3d-pi4.dtbo \
+    overlays/vc4-kms-dsi-7inch.dtbo \
     overlays/vc4-kms-v3d.dtbo \
     overlays/vc4-kms-v3d-pi4.dtbo \
-    overlays/vc4-kms-dsi-7inch.dtbo \
     overlays/w1-gpio.dtbo \
     overlays/w1-gpio-pullup.dtbo \
     overlays/wm8960-soundcard.dtbo \
--- a/docs/extra-build-config.md
+++ b/docs/extra-build-config.md
@@ -108,13 +108,6 @@ Example to force HDMI output to 720p in CEA mode:
 
 See: <https://www.raspberrypi.com/documentation/computers/configuration.html#hdmi-configuration>
 
-## Video camera support with V4L2 drivers
-
-Set this variable to enable support for the video camera (Linux 3.12.4+
-required):
-
-    VIDEO_CAMERA = "1"
-
 ## Enable offline compositing support
 
 Set this variable to enable support for dispmanx offline compositing:
@@ -389,28 +382,25 @@ option:
         # Raspberry Pi 7\" display/touch screen \n \
         lcd_rotate=2 \n \
         '
-## Enable Raspberry Pi Camera Module
+## Camera Modules
 
-Raspberry Pi does not have the unicam device ( Raspberry Pi Camera ) enabled by default.
-Because this unicam device ( bcm2835-unicam ) as of now is used by libcamera opensource.
-So we have to explicitly enable it in local.conf.
+Automatic camera detection is enabled by default, which is equivalent to:
 
-    RASPBERRYPI_CAMERA_V2 = "1"
+    RASPBERRYPI_CAMERA = "auto"
 
-This will add the device tree overlay imx219 ( Raspberry Pi Camera Module V2 sensor driver 
-) to config.txt. Also, this will enable adding Contiguous Memory Allocation value in the 
-cmdline.txt.
+To override the automatic camera detection, clear `RASPBERRYPI_CAMERA`
+or specify an explicit camera module name. Common values are predefined to
+avoid case sensitivity issues or typos, like `${RASPBERRYPI_CAMERA_V2}`.
+For example, request the support for the Camera Module 3 with either:
 
-Similarly, the Raspberry Pi Camera Module v3 also has to be explicitly enabled in local.conf.
+    RASPBERRYPI_CAMERA = "${RASPBERRYPI_CAMERA_V3}"
 
-    RASPBERRYPI_CAMERA_V3 = "1"
+or directly:
 
-This will add the device tree overlay imx708 ( Raspberry Pi Camera Module V3 sensor driver ) 
-to config.txt.
+    RASPBERRYPI_CAMERA = "imx708"
 
 See:
-* <https://www.raspberrypi.com/documentation/computers/camera_software.html>
-* <https://www.raspberrypi.org/blog/an-open-source-camera-stack-for-raspberry-pi-using-libcamera/>
+* <https://www.raspberrypi.com/documentation/computers/camera_software.html#if-you-do-need-to-alter-the-configuration>
 
 ## WM8960 soundcard support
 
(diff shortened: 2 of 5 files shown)


vivien:
This PR also fixes #1177 and #1197.

conf/machine/include/rpi-base.inc
```diff
@@ -31,6 +31,10 @@ RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
     overlays/i2c-gpio.dtbo \
     overlays/i2c-rtc.dtbo \
     overlays/imx219.dtbo \
+    overlays/imx290.dtbo \
```
agherzan, line 34: Does it work on all the kernel versions we currently have in the layer?

recipes-bsp/bootfiles/rpi-config_git.bb
```diff
@@ -45,84 +45,84 @@ do_deploy() {
     cp ${S}/config.txt $CONFIG
 
     if [ -n "${KEY_DECODE_MPG2}" ]; then
-        sed -i '/#decode_MPG2=/ c\decode_MPG2=${KEY_DECODE_MPG2}' $CONFIG
+        echo "decode_MPG2=${KEY_DECODE_MPG2}" >> $CONFIG
```
agherzan, line 52: I understand what this does, but the annoyance is that we move context from the sections that we have in the config. Maybe we should have a sed that changes existing lines **or** appends when it isn't found. Maybe an awk?

recipes-bsp/bootfiles/rpi-config.bb
```diff
@@ -2,18 +2,16 @@ DESCRIPTION = "Commented config.txt file for the Raspberry Pi. \
                The Raspberry Pi config.txt file is read by the GPU before \
                the ARM core is initialised. It can be used to set various \
                system configuration parameters."
+HOMEPAGE = "https://www.raspberrypi.com/documentation/computers/config_txt.html"
```
agherzan, line 5: I'm not sure what to say about this commit. I feel like users are used to the format and will end up with a pretty confusing change. @kraj what do you think?

recipes-bsp/bootfiles/rpi-config.bb
```diff
@@ -209,24 +230,6 @@ do_compile() {
         echo "dtoverlay=${VC4DTBO}" >> $CONFIG
     fi
 
-    # Choose Camera Sensor to be used, default imx219 sensor
-    if [ "${RASPBERRYPI_CAMERA_V2}" = "1" ]; then
```
agherzan, line 216: We should make this backwards compatible.

agherzan: At least for a while with a warning.

agherzan (changes requested):

conf/machine/include/rpi-base.inc
vivien: It should, why wouldn't it?

recipes-bsp/bootfiles/rpi-config_git.bb
vivien: There's no need to force a context whatsoever. If there is an issue with the ordering (like the start_x parameter), this can be explicitly documented, but there is no need to provide arbitrary context in the meantime.

recipes-bsp/bootfiles/rpi-config.bb
vivien: I really don't think that a commented out file was adding any value to the user. On the contrary, this was just cluttered and error prone. The only useful addition to the config file would be to echo the homepage URL as the first line so that the user knows where to look for documentation if they want to tweak the config, not scrolling a subjectively commented out file.

recipes-bsp/bootfiles/rpi-config.bb
vivien: Yes we can add something like:

```
if [ "${RASPBERRYPI_CAMERA_V2}" = "1" ]; then
    # warn...
    RASPBERRYPI_CAMERA="imx..."
fi
```

recipes-bsp/bootfiles/rpi-config_git.bb
agherzan: Unless you have been using this layer for years and are relying on that information and format.

recipes-bsp/bootfiles/rpi-config.bb
agherzan: I see both sides - I'm unsure how to proceed.;

agherzan: @kraj your thoughts on this?

recipes-bsp/bootfiles/rpi-config_git.bb
agherzan: @kraj your thoughts on this?

recipes-bsp/bootfiles/rpi-config.bb
```diff
@@ -36,6 +36,13 @@ GPIO_SHUTDOWN_PIN ??= ""
 
 inherit deploy nopackages
 
+RASPBERRYPI_CAMERA ??= "auto"
+RASPBERRYPI_CAMERA_V1 = "ov5647"
+RASPBERRYPI_CAMERA_V2 ?= "imx219"
+RASPBERRYPI_CAMERA_HQ = "imx477"
+RASPBERRYPI_CAMERA_GS = "imx296"
+RASPBERRYPI_CAMERA_V3 ?= "imx708"
```
agherzan, line 44: Any reason to have a mix of assignments and weak-assignments here?

agherzan: BTW, I like this approach.

recipes-bsp/bootfiles/rpi-config.bb
```diff
@@ -36,6 +36,13 @@ GPIO_SHUTDOWN_PIN ??= ""
 
 inherit deploy nopackages
 
+RASPBERRYPI_CAMERA ??= "auto"
+RASPBERRYPI_CAMERA_V1 = "ov5647"
+RASPBERRYPI_CAMERA_V2 ?= "imx219"
+RASPBERRYPI_CAMERA_HQ = "imx477"
```
agherzan, line 42: What uses some of these variables?

agherzan (changes requested):

recipes-bsp/bootfiles/rpi-config_git.bb
kraj: We have run into problems in past when we try to add lines, it could be that bootloader parser is limited in what it can read. I remember one pull where a setting would not take effect because line length was getting longer than 80 chars or some such. So I am a bit nervious to replace sed with echo like this everywhere.

recipes-bsp/bootfiles/rpi-config.bb
vivien: The weak assignment is to not override the retro compatibility with the previous RASPBERRYPI_CAMERA_V2/V3 = "1" flag and issue a warning instead.

recipes-bsp/bootfiles/rpi-config.bb
vivien: These variables describe the common values and are optionally used by the user to assign RASPBERRYPI_CAMERA if they wish not to explicit the actual model name.

agherzan: Can you give me an example?

recipes-bsp/bootfiles/rpi-config.bb
agherzan: @vivien can you add a code comment for that? It will be very easy to miss in the future.

recipes-bsp/bootfiles/rpi-config.bb
vivien: `RASPBERRYPI_CAMERA = "${RASPBERRYPI_CAMERA_V2}"`, it's optional, but convenient in order to avoid case sensitive errors or other typos. I'll update the doc to reflect that as well.

recipes-bsp/bootfiles/rpi-config_git.bb
vivien: > We have run into problems in past when we try to add lines, it could be that bootloader parser is limited in what it can read. I remember one pull where a setting would not take effect because line length was getting longer than 80 chars or some such. So I am a bit nervious to replace sed with echo like this everywhere.

sed doesn't guarantee anything. This is already handled and actually improved in 24984b6... see the commit message.

recipes-bsp/bootfiles/rpi-config.bb
kraj: config.txt has detailed comments for an option, in-place sed keeps that intact. Non yocto users are familiar with this as well, since thats the first place people go to debug weird issues. So keeping the format has advantages. Sed is as reliable as echo, both are shell commands. They can fail in same ways. If sed can go wrong so can echo emit an older option and render system problems. So I am not buying echo is better than sed argument :)

kraj: how do we handle the fact that our config.txt mechanism is our of sync with upstream config.txt, does this happen ?
do we then point them to wrong config.txt as reference ? Secondly, I think lot of users edit it in place to debug things so a symlink at top might be fine to point them to what these options mean maybe

vivien: I'm not sure to understand what you mean by upstream config.txt, reference config.txt or symlink. I also don't understand how a file fetched from some random user repository on Github tackles any of the issues you're trying to describe here. The cleaner the file is, the less error-prone it gets.

kraj: the original `config.txt` that we will overwrite if we accept this PR.

vivien: I want to clarify that there isn't an "original" config.txt file that we are overwriting with this PR. The initial recipe fetched an [unofficial file](https://github.com/Evilpaul/RPi-config) from GitHub.

Different distributions come with their own config.txt files. For instance, RPI-Distro (formerly Raspbian) ships [this one](https://raw.githubusercontent.com/RPi-Distro/pi-gen/master/stage1/00-boot-files/files/config.txt), while Arch Linux ARM ships [this version](https://github.com/archlinuxarm/PKGBUILDs/blob/master/core/linux-rpi/config.txt). The choice of config.txt is determined by the distribution, not the recipe.

In fact, a minimal image can successfully boot with either no config.txt file or with the following content:

```
# Empty file
# Some settings may impact device functionality. Refer to the official documentation below for details.
# https://www.raspberrypi.com/documentation/computers/config_txt.html
```

As a BSP layer, our goal should be to provide a clean default configuration with minimal adjustments. Distributions have the flexibility to further customize the configuration using conditional filters, additional files like autoboot.txt, or any configuration fragments enabled with the include directive, as outlined in the documentation.

Following best practices, it would be advisable to list rpi-config in MACHINE_EXTRA_RDEPENDS to prevent core-image-minimal from installing the file by default, while allowing other core images to include it. In the meantime, I kindly request that @kraj and @agherzan carefully consider this comment and refrain from advocating for the use of an unofficial text file on the internet, as it may not align with established best practices.

vivien:
See this [comment](https://github.com/agherzan/meta-raspberrypi/pull/1213#discussion_r1337327503) for explanation behind the motivation for a clean config file.

conf/machine/include/rpi-base.inc
agherzan: So, do they exist in all the kernel versions master currently provides?

recipes-bsp/bootfiles/rpi-config_git.bb
agherzan: To not block this, what about splitting this into a separate PR?

recipes-bsp/bootfiles/rpi-config.bb
agherzan: I would recommend this being split in another PR. It does not directly support the camera modules and we can have more public visibility on this in this way.

recipes-bsp/bootfiles/rpi-config.bb
agherzan: I propose keeping this simple with only RASPBERRYPI_CAMERA and giving documentation examples of its values. Having indirect definitions and keeping them updated with new variables is a situation that might cause more confusion.

docs/extra-build-config.md
```diff
@@ -389,28 +382,25 @@ option:
...
-This will add the device tree overlay imx219 ( Raspberry Pi Camera Module V2 sensor driver 
-) to config.txt. Also, this will enable adding Contiguous Memory Allocation value in the 
-cmdline.txt.
+To override the automatic camera detection, clear `RASPBERRYPI_CAMERA`
+or specify an explicit camera module name. Common values are predefined to
+avoid case sensitivity issues or typos, like `${RASPBERRYPI_CAMERA_V2}`.
+For example, request the support for the Camera Module 3 with either:
 
-Similarly, the Raspberry Pi Camera Module v3 also has to be explicitly enabled in local.conf.
+    RASPBERRYPI_CAMERA = "${RASPBERRYPI_CAMERA_V3}"
```
agherzan, line 396: There are more values possible here. I would only keep RASPBERRYPI_CAMERA and have examples of direct assignments instead of having to document indirect variables.

agherzan (changes requested):

recipes-bsp/bootfiles/rpi-config.bb
djr-spectrummfg: I would personally prefer a minimal config.txt that isn't filled with every option imaginable.

agherzan: @vivien I started to lean into that, too - with the reoccurrence of the config.txt size issues in RPI5. @kraj I think we should start considering this as a cleaner solution forward.

agherzan:

PR #1214, opened 2023-08-15
Install object_detect_tf post processing stage

vivien:
With this PR, one can add the [meta-neural-network](https://github.com/nnstreamer/meta-neural-network) layer to their system, enable tensorflow support in libcamera-apps with e.g.:
```
PACKAGECONFIG:pn-libcamera-apps:append = " tflite"
```
 and try object detection with the new libcamera-detect app or even `libcamera-still --post-process-file /usr/share/libcamera-apps/assets/object_detect_tf.json --lores-width 400 --lores-height 300` for example.
 
 (tested on raspberrypi4-64 with a Camera Module 3)

### DIFF
--- a/dynamic-layers/neuralnetwork/recipes-tensorflow/tensorflow-lite/tensorflow-lite_2.%.bbappend
+++ b/dynamic-layers/neuralnetwork/recipes-tensorflow/tensorflow-lite/tensorflow-lite_2.%.bbappend
@@ -1,3 +1,6 @@
+# This provider of tensorflow-lite only builds on 64-bit systems
+COMPATIBLE_MACHINE:rpi = "^(aarch64)$"
+
 # https://github.com/nnstreamer/meta-neural-network/pull/85
 do_install:append () {
 	ln -sf libtensorflow2-lite.a ${D}${libdir}/libtensorflow-lite.a
--- a/dynamic-layers/neuralnetwork/recipes-tensorflow/tensorflow-lite/tensorflow-lite_2.%.bbappend
+++ b/dynamic-layers/neuralnetwork/recipes-tensorflow/tensorflow-lite/tensorflow-lite_2.%.bbappend
@@ -1,0 +1,5 @@
+# https://github.com/nnstreamer/meta-neural-network/pull/85
+do_install:append () {
+	ln -sf libtensorflow2-lite.a ${D}${libdir}/libtensorflow-lite.a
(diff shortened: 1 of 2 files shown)


vivien:
tensorflow-lite from meta-neural-network currently only builds for 64-bit systems, please let me know the best way to skip the build for 32-bit systems (libcamera-apps will build just fine if tensorflow was requested by the user but not available in sysroot.)

dynamic-layers/neuralnetwork/recipes-tensorflow/tensorflow-lite/tensorflow-lite_2.%.bbappend
```diff
@@ -0,0 +1,4 @@
+do_install:append () {
```
agherzan, line 1: I still find this an issue that needs to be fixed in the actual layer. I'm not sure I understand the reasoning not to do that.

agherzan (changes requested):

dynamic-layers/neuralnetwork/recipes-tensorflow/tensorflow-lite/tensorflow-lite_2.%.bbappend
kraj: right this is a generic change.

vivien: Obviously that is a generic change that is being discussed [upstream](https://github.com/nnstreamer/meta-neural-network/pull/85). The reasoning behind having a bbappend with the fix in our layer is to document the issue in the meantime, while still providing a functional example to the users. I'm not sure if you noticed, but processing with tensorflow did not work until now.

agherzan: I see that the review is in progress, and it introduces some compile issues. Let's see how that settles upstream (see the symlink discussion). I want to avoid:
1. introducing new issues while fixing this
2. making sure that this gets cleaned up after upstream deals with it

vivien:
I've documented the upstream issue in the code and commit message and skipped the build on 32-bit systems (libcamera-apps builds fine when tflite isn't available in sysroot.)

agherzan:

PR #1220, opened 2023-09-08
Update kernel/firmaware 

kraj:
**- What I did**

**- How I did it**


recipes-kernel/linux-firmware-rpidistro/linux-firmware-rpidistro_git.bb
```diff
@@ -47,19 +47,20 @@ do_install() {
 
     # add compat links. Fixes errors like
     # brcmfmac mmc1:0001:1: Direct firmware load for brcm/brcmfmac43455-sdio.raspberrypi,4-model-compute-module.txt failed with error -2
-    ln -s brcmfmac43455-sdio.txt ${D}${nonarch_base_libdir}/firmware/brcm/brcmfmac43455-sdio.raspberrypi,4-compute-module.txt
+    #ln -s brcmfmac43455-sdio.txt ${D}${nonarch_base_libdir}/firmware/brcm/brcmfmac43455-sdio.raspberrypi,4-compute-module.txt
```
agherzan, line 50: What's up with these comments?

kraj: These symlinks are already part of new prebuilt. I have kept them in comments for reference.

agherzan: That sounds great! In that case I would propose to drop these from the recipe and reference them in it log. In this way the info remains easy to access while we don't have to maintain comments and references in the recipe. What do you think?

kraj: done

agherzan: I hate to be nit picker but we missed a comment in git log for this drop.

kraj: indeed I did. Fixed now.

agherzan (commented):
Thanks!

mdsketch:
@kraj I can't seem to find much information on what the latest version of the RPI firmware is. I see some tags, in some repos, but I can't seem to figure it out.

I was wondering if you could help me to understand why you are updating to the buster version of the firmware and not the latest version of bullseye?

It is my understanding that if we are on the 6.1.54 kernel for the yocto recipe, we would want to also have the bullseye firmware as it will support that kernel (as opposed to 5.x.x version for buster).

I have tried the latest bullseye and it works just fine. Should I expect issues? or should I be fine?

https://github.com/RPi-Distro/firmware/blob/pios/bullseye/debian/changelog
https://github.com/raspberrypi/firmware/releases/tag/1.20230405

kraj:
You should be fine if it tests out fine infact post a pull here as well

mdsketch:
Sounds good thanks, we've been testing quite a bit if don't see any issues I might open a PR.

PR #1238, opened 2023-10-25
Kirkstone rpi-eeprom 

djr-spectrummfg:
This is a simple backport of the changes from #1131 to the kirkstone branch.


agherzan (changes requested):
You included a vim commit - probably by mistake.

recipes-support/vim/vim_9.0.bbappend
```diff
@@ -0,0 +1 @@
+RPROVIDES:${PN} += " xxd"
```
holagvk, line 1: I think instead of using `vim` to provide `xxd`, we can use `xxd` from `busybox`?

1) `busybox_%.bbappend`:
```
FILESEXTRAPATHS:append := ":${THISDIR}/${PN}"
SRC_URI += "file://enable-xxd.cfg"
RPROVIDES:${PN} += " xxd"
```

2) Add `enable-xxd.cfg`
`CONFIG_XXD=y`

djr-spectrummfg: Using vim is what the master branch does, so the RPROVIDES="xxd" change was backported from that. I wasn't aware busybox provided an implementation of xxd either.

djr-spectrummfg:
> You included a vim commit - probably by mistake.

This is not a mistake. It is required for it to work and this is a backport from master.

recipes-support/vim/vim_9.0.bbappend
agherzan: Sure, but this is just not right in the BSP without any additional guards, explanations etc.

agherzan (changes requested):

recipes-support/vim/vim_9.0.bbappend
kraj: we already use `update-alternatives` to provide appropriate xxd. If we do not include `vim-xxd` in image then it will nicely fallback to busybox to provide it. We perhaps do not require anything to do here.

djr-spectrummfg: I don't know what to tell you other than if you remove that it doesn't build.

holagvk: I confirmed as well - Without `vim` or `busbybox` explicitly providing `xxd`, `rpi-eeprom` doesn't build.

agherzan: How does this work in the source of the backport? We really need a way to first understand the problem. All we have is a commit that says `Add RPROVIDES="xxd" to vim so that rpi-eeprom can be built on kirkstone branch`. We need an explanation (not to mention the commit message format (https://meta-raspberrypi.readthedocs.io/en/latest/contributing.html),

agherzan:

djr-spectrummfg:
It's abandoned because you refused to accept the working solution that is a direct backport.

recipes-support/vim/vim_9.0.bbappend
kraj: I wonder if https://lore.kernel.org/openembedded-core/20241204230907.43936-1-denis@denix.org/ helps fix it ?

djr-spectrummfg: The end result will be the same: vim will be installed as a dependency

PR #1239, opened 2023-10-25
rpi-eeprom_git: Update recipe to latest rpi-eeprom repo

mdsketch:
I needed newer versions of the firmware so I updated this recipe and I was wondering if it's something we can update here as well? Wasn't sure if I should make a PR or Issue, so here we go:

**- What I did**

I upped the commit version to the latest release of https://github.com/raspberrypi/rpi-eeprom

This will allow updating the pi4 to the latest version eeprom. It does not support the pi5 (but it could with some more editing).

**- How I did it**

I had to change the directories that are installed as they changed upstream. I matched the compatibility that upstream provides.

Also removed the patch as it's already changed.


agherzan (changes requested):
Hello,

Thanks for this. There are some things that we need to do before we can go ahead with this PR.
1. Changes firstly go to master as a PR, and once that is done, you can raise another PR with a backport as a set of clean cherry-picks.
2. git commits need to follow the existing guidelines and history - https://meta-raspberrypi.readthedocs.io/en/latest/contributing.html
3. I'm not really sure what is happening with the removed patch? Was it included upstream? If yes, let's squash it in one commit with a relevant git log mention.

mdsketch:
Ok I'll make those changes. And yes as far as I can tell the patch was included upstream, I have tested this change on my Mickledore branch so that's why I was merging it here, but I will change that.

Just so we have the record in this PR, here is the upstream change that does the same thing as the patch https://github.com/raspberrypi/rpi-eeprom/blob/5ec5c003bacc73847aadad712aa1fbdace8f1c4e/rpi-eeprom-update#L122

and here is the actual patch commit https://github.com/raspberrypi/rpi-eeprom/commit/55ca589ab9e6e719c1e1f9fc72997d1643d28fe3

PR #1241, opened 2023-10-27
Update raspberrypi-firmware to 20231027 to support CM4 1.1 and Pi3B+ 1.4

### LINKED ISSUE #1173: dunfell firmware files too old for newer revisions of 3B+
https://github.com/agherzan/meta-raspberrypi/blob/dunfell/recipes-bsp/common/raspberrypi-firmware.inc

https://pip.raspberrypi.com/categories/797-pcn

The firmware seems to be dated before this PMIC change and the 3B+ will not boot (4 long flashes  followed by 7 short flashes)

ColinFinck:
Fixes #1173

### DIFF
--- a/recipes-bsp/common/raspberrypi-firmware.inc
+++ b/recipes-bsp/common/raspberrypi-firmware.inc
@@ -1,9 +1,9 @@
-RPIFW_DATE ?= "20210527"
-SRCREV ?= "1eb8921ae7158249894f89e8bb6c4bfb04ec7377"
+RPIFW_DATE ?= "20231027"
+SRCREV ?= "9814645c19bd8621fdd382e13369b1efea816c1c"
 RPIFW_SRC_URI ?= "https://github.com/raspberrypi/firmware/archive/${SRCREV}.tar.gz;downloadfilename=raspberrypi-firmware-${SRCREV}.tar.gz"
 RPIFW_S ?= "${WORKDIR}/firmware-${SRCREV}"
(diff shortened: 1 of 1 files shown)


ColinFinck:
CC @agherzan

agherzan:
This is a tricky one, as you might expect. We are bumping a pretty fundamental package in the BSP in a long-term supported branch. Your commit log needs to be a rationale about this enough to make a decision and document the possible impact and known gains. Also - please take a look at https://meta-raspberrypi.readthedocs.io/en/latest/contributing.html and restructure your commit a bit.

agherzan:
I'll close this PR for now but feel free to comment and we can resurrect it if needed. Thanks.

Awarty:
I know this issue has been closed for almost a year but I wanted to add some information to this problem. I ran into an issue where new CM4:ors would complain about the firmware of the image created using Yocto and this meta layer for Raspberry Pi. I used the Dunfell version of Yocto and the firmware version used by the Dunfell branch of meta-raspberrypi points to a too-old version for newer CM4:ors. To solve our issue I used the fix suggested by @ColinFinck  by adding bootfiles.bbappend to override the variables. After testing on old and new CM4 devices, this update solved my issues with the firmware version. This fix would be useful if it could be added to the Dunfell branch but for now, my fix solves the issue for me. Just wanted to add information that this "problem" affected another user as well and that the fix could be something worth looking into.

PR #1242, opened 2023-11-08
changes done in meta-raspberrypi

Besquarelibrary:
**- What I did**

**- How I did it**


recipes-kernel/linux/linux-raspberrypi_5.4.bb
```diff
@@ -11,4 +11,5 @@ SRC_URI += "file://0001-Revert-selftests-bpf-Skip-perf-hw-events-test-if-the.pat
             file://0002-Revert-selftests-bpf-Fix-perf_buffer-test-on-systems.patch \
             file://powersave.cfg \
             file://android-drivers.cfg \
+	    file://my_config.cfg \
```
kraj, line 14: The file name is not appropriate perhaps it could be a tft.conf or add specific tft name to qualify it more

kraj (changes requested):

agherzan (changes requested):
Hello @Besquarelibrary, and welcome to this project. We have some guidelines in https://meta-raspberrypi.readthedocs.io/en/latest/contributing.html that should give you enough context on how to put up a PR for meta-raspberrypi. You can always check git history for examples because we want to maintain a clean and useful history.

Also, remember that changes need to land in master before they can be proposed for a backport in a stable branch. Of course, there are exceptions, but they need to be clearly documented and agreed upon.

recipes-kernel/linux/linux-raspberrypi_5.4.bb
agherzan: I suspect this is for piscreen.

agherzan:
@Besquarelibrary I will close this for now while you clean up and experiment with your changes. Once you have things polished, let's put up a master PR.

PR #1245, opened 2023-11-10
linux-raspberrypi_6.1.bb: Update to 6.1.61 release

kraj:


agherzan (approved):
Thanks. Approved and merged.

agherzan:
kirkstone doesn't even have a 6.X version, and changing a major kernel version in a stable branch is not something that I would jump onto without some deep considerations on regressions. Is mixin layer something we could work with as a workaround here?

I can't speak of other branches as this needs to be dealt with on a case-by-case basis.

floion:
Kirkstone being the LTS release supported until 2026 I think it would make sense to have kernel updates on that branch, especially since we'd need a kernel update to support the Pi 5.
What do you mean about mixing layers?

agherzan:
I was thinking of something like: https://docs.yoctoproject.org/ref-manual/terms.html#term-Mixin
Here is an example: https://git.yoctoproject.org/meta-lts-mixins

I could create a `meta-raspberrypi-mixin` where we could backport (community driven) this kind of changes.

floion:
If you never plan on updating the kernel version in a LTS release then I guess we could try mixins.

floion:
@agherzan you can hold on that for the moment, I don't think we'll go the mixin path after all.

floion:
Can this PR instead be ported to nanbield? Or mickledore if nanbield is not supported by meta-raspberrypi outside of master?

agherzan:
I am okay with a backport on supported Yocto versions as long as it is a patch update as this MR.

PR #1237, opened 2023-10-25
Add Raspberry Pi 5

floion:
**- What I did**

**- How I did it**

PR #1247, opened 2023-11-17
rpi-config: Upgrade to tip of tree

leon-anavi:
Upgrade to the latest version of the repository with includes the following improvements of the template for config.txt:

- update revision and date stamp
- additon of dtoverlay
- general updates

This work was sponsored by GOVCERT.LU.

**- What I did**

Upgrade to the latest version of https://github.com/Evilpaul/RPi-config

**- How I did it**

While working on https://github.com/agherzan/meta-raspberrypi/pull/1237 I noticed that recipe `rpi-config` uses an older version and decided to upgrade the template for `config.txt` to the latest available version. The primary reason is that in the new version the date at the beginning of `config.txt` has been changed for `Revision 16, 2013/06/22` to `Revision 17, 2021/08/15`. Obviously there have been many additions in general to `config.txt` since 2013 so it is better to use a newer date.

### DIFF
--- a/recipes-bsp/bootfiles/rpi-config_git.bb
+++ b/recipes-bsp/bootfiles/rpi-config_git.bb
@@ -7,7 +7,7 @@ LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda
 
 COMPATIBLE_MACHINE = "^rpi$"
 
-SRCREV = "648ffc470824c43eb0d16c485f4c24816b32cd6f"
+SRCREV = "6ac2d832c6c3b208e2669f50ec1abf2c20cb7ff4"
 SRC_URI = "git://github.com/Evilpaul/RPi-config.git;protocol=https;branch=master \
           "
 


holagvk:
@agherzan / @leon-anavi The latest revision of `rpi-config` removes a lot of configurations (e.g. `start_x`) in this patch: https://github.com/Evilpaul/RPi-config/commit/78181eee761983c5c920288dca7a32bd9fd61ed4

So, this PR will break `meta-raspberrypi` configurations in https://github.com/agherzan/meta-raspberrypi/blob/master/recipes-bsp/bootfiles/rpi-config_git.bb  (e.g., VIDEO_CAMERA will no longer be able to set `start_x=1`)

I think we should revert this commit https://github.com/agherzan/meta-raspberrypi/commit/60082497227aad169407137c651a214321b53b66 ?

leon-anavi:
@holagvk good point. Wouldn't it be better to identify the missing configurations that are still needed and add them to https://github.com/Evilpaul/RPi-config ?

On Monday I will try to find them by comparing `do_deploy` of `rpi-config_git.bb` to the current default `config.txt` from https://github.com/Evilpaul/RPi-config.

leon-anavi:
@holagvk, I checked all `sed` commands from `rpi-config.bb` against the latest `config.txt` template. I wasn't able to identify any other missing configurations apart from the one you found. So I created a GitHub pull request to reintroduce `start_x` to the `config.txt` template: https://github.com/Evilpaul/RPi-config/pull/8

agherzan:
This shows an issue with this approach - we rely on upstream content for our configuration.

PR #1166, opened 2023-04-04
machines: simplify MACHINEOVERRIDES definitions

zach-welch-aquabyte:
Prepending new definitions should be preferred to assignment, as it is simpler and more conducive to defining new machine configurations that reuse these configurations.

I posted an initial draft of this patch (with the required `Signed-Off-By` line) to the yocto mailing list as well, but I thought a PR might elicit a more expedient response.

Also, I tested this on the `kirkstone` branch with the `raspberrypi{3,4}-64` machines included in my own custom machine configurations, but these changes should apply cleanly on all active branches and be safe to use with all configurations.

PR #1248, opened 2023-11-18
machines: simplify MACHINEOVERRIDES definitions

shr-project:
Prepending new definitions should be preferred to assignment, as it is simpler and more conducive to defining new machine configurations that reuse these configurations.


shr-project:
This is just to sync mickledore with what was already merged in kirkstone in:
https://github.com/agherzan/meta-raspberrypi/pull/1166
and master:
https://github.com/agherzan/meta-raspberrypi/pull/1179

agherzan (approved):
Thanks.

PR #1252, opened 2023-11-27
docs: fix syntax for overriding fs type for initramfs image

lorenzo-arena:
**- What I did**

I fixed the documentation for initramfs: I tried creating an initramfs image following the official (latest) documentation of the meta-raspberrypi layer but there was a dependency loop error caused by the `IMAGE_FSTYPES_pn-${INITRAMFS_IMAGE}` variable not following the latest syntax for overrides.

**- How I did it**

I fixed the doc by using the `IMAGE_FSTYPES:pn-${INITRAMFS_IMAGE}` syntax for the override.


agherzan (approved):
Thanks!

PR #1261, opened 2024-01-05
psplash: Add udev rule notify systemd of available fb

Kiwi4Inovex:
When using psplash in combination with systemd the splash screen is not shown. The dependency to sys-devices-platform-gpu-graphics-fb0.device will terminate psplash-start.service because systemd is not aware of the framebuffer device node.
See https://lists.yoctoproject.org/g/yocto/topic/91286438#57156

Adding a udev rule to notify systemd does the trick


agherzan:
I suspect that this is not Mickledore-specific, so it should go to master first.

Kiwi4Inovex:
Unfortunately i can't change the source branch, i reopen another PR.

PR #1266, opened 2024-01-18
Define CAN interrupt pins

### LINKED ISSUE #1264: Define CAN interrupt pin as a variable
**Description**

It seems that the interrupt pin used for CAN is hard-coded to be 25. I suggest to turn this into a variable in the same way as it is for CAN_OSCILLATOR.

I'm referring to [this line of code](https://github.com/agherzan/meta-raspberrypi/blob/8bad47eba3a56204a5b5d2a433b499398bda2173/recipes-bsp/bootfiles/rpi-config_git.bb#L255C90-L255C90)

dferrarigrowtech:
Fixes #1264

**- What I did**
The users can now define CAN interrupt pins that are different from the hard-coded one.

**- How I did it**
I defined two more CAN variables that let users configure the CAN interrupt pins depending by the module that they used (or by their custom boards). It is similar to CAN_OSCILLATOR. I also added these variables to the documentation

Once this is merged I will cherry-pick this commit to Kirkstone (because I need this change)

### DIFF
--- a/docs/extra-build-config.md
+++ b/docs/extra-build-config.md
@@ -340,6 +340,13 @@ Some modules may require setting the frequency of the crystal oscillator used on
 
     CAN_OSCILLATOR="8000000"
 
+Configure the interrupt pin to the one connected to the CAN module. By default,
+the pins are set to 25 for can0 and 24 for can1. To change them to 12 and 16,
+the following variables also have to be set:
+
+    CAN0_INTERRUPT_PIN = "12"
+    CAN1_INTERRUPT_PIN = "16"
+
 Tested modules:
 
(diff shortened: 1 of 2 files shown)


agherzan (changes requested):
Thanks! This looks good content-wise but it lacks adhering to the guidelines in https://meta-raspberrypi.readthedocs.io/en/latest/contributing.html. Could you have a short check on it and refresh this?

dferrarigrowtech:
Thank you @agherzan! I'm not used to contributions with mail patches.
What is exactly the issue? Should I change the commit message or send the commit via email?

dferrarigrowtech:
Thank you @agherzan. I hope that now it is ok

agherzan (approved):
Thanks! Looks good.

PR #1267, opened 2024-01-26
[Kirkstone] Backport of CAN interrupt pin patch

dferrarigrowtech:
Just a cherry picking of the two commits merged yesterday about CAN interrupt pins.


dferrarigrowtech:
I don't know if I should also cherry pick the commit about the documentation. I tried but the CI failed, so I removed that commit. Waiting for instructions

agherzan:
I would pull in the documentation as well.

agherzan (approved):

dferrarigrowtech:
@agherzan all the checks have passed except the docs one. I noticed that check was not passed even before I added the docs commit. What could it be? Can we merge this anyway?

PR #1268, opened 2024-01-26
psplash: udev rule to notify systemd of available framebuffers

Kiwi4Inovex:
When using psplash in combination with systemd the splash screen is not shown. The dependency to sys-devices-platform-gpu-graphics-fb0.device will terminate psplash-start.service because systemd is not aware of the framebuffer device node.
See https://lists.yoctoproject.org/g/yocto/topic/91286438#57156

Adding a udev rule to notify systemd

### DIFF
--- a/recipes-core/udev/udev-rules-rpi.bb
+++ b/recipes-core/udev/udev-rules-rpi.bb
@@ -5,6 +5,7 @@ LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda
 SRC_URI = " \
 	git://github.com/RPi-Distro/raspberrypi-sys-mods;protocol=https;branch=master \
 	file://can.rules \
+	file://fb.rules \
 	"
 SRCREV = "5ce3ef2b7f377c23fea440ca9df0e30f3f8447cf"
 
@@ -16,4 +17,5 @@ do_install () {
     install -d ${D}${sysconfdir}/udev/rules.d
     install -m 0644 ${S}/etc.armhf/udev/rules.d/99-com.rules ${D}${sysconfdir}/udev/rules.d/
     install -m 0644 ${UNPACKDIR}/can.rules ${D}${sysconfdir}/udev/rules.d/
+    install -m 0644 ${UNPACKDIR}/fb.rules ${D}${sysconfdir}/udev/rules.d/
 }
--- a/recipes-core/udev/udev-rules-rpi/fb.rules
+++ b/recipes-core/udev/udev-rules-rpi/fb.rules
@@ -1,0 +1,1 @@
+SUBSYSTEM=="graphics", KERNEL=="fb[0-9]*", TAG+="systemd"


Kiwi4Inovex:
Follow up of https://github.com/agherzan/meta-raspberrypi/pull/1261 based on a new branch

recipes-core/psplash/psplash_%.bbappend
```diff
@@ -1,11 +1,16 @@
...
+SRC_URI:append:rpi = " \
+    file://framebuf.conf \
+    file://fb.rules \
+    "
 
 do_install:append:rpi() {
     if [ "${@bb.utils.filter('DISTRO_FEATURES', 'systemd', d)}" ]; then
         install -Dm 0644 ${WORKDIR}/framebuf.conf ${D}${systemd_system_unitdir}/psplash-start.service.d/framebuf.conf
+        install -d ${D}${sysconfdir}/udev/rules.d
+        install -m 0644 ${WORKDIR}/fb.rules ${D}${sysconfdir}/udev/rules.d/
```
kraj, line 13: Reduce from two lines to a single line with
`install -Dm 0644 ${WORKDIR}/fb.rules ${D}${sysconfdir}/udev/rules.d/fb.rules`

recipes-core/psplash/psplash_%.bbappend
```diff
@@ -1,11 +1,15 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
 SPLASH_IMAGES:rpi = "file://psplash-raspberrypi-img.h;outsuffix=raspberrypi"
 
-SRC_URI:append:rpi = " file://framebuf.conf"
```
agherzan, line 4: Is there any reason not to include this in the existing rules conf file?

Kiwi4Inovex: It's bound to the decision if systemd is activated. From my point it belongs in the same decision tree.

kraj: the rule has runtime check for systemd so it can go in common rules, however, fb0 might not be same name for framebuffer on all devices

Kiwi4Inovex: Applied. Note the framebuf.conf is pointing to /sys/devices/platform/gpu/graphics/fb0

kraj (approved):
lgtm, without this patch we dont get splash with systemd. So we need this patch

kraj:
@Kiwi4Inovex can you rebase your branch on top of latest master and push this again so we can merge it with clean history.

recipes-core/udev/udev-rules-rpi/fb.rules
```diff
@@ -0,0 +1 @@
+SUBSYSTEM=="graphics", KERNEL=="fb[0-9]*", TAG+="systemd"
```
agherzan, line 1: Small request: Include a newline at the end of the last line.

recipes-core/udev/udev-rules-rpi.bb
```diff
@@ -5,6 +5,7 @@ LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda
 SRC_URI = " \
 	git://github.com/RPi-Distro/raspberrypi-sys-mods;protocol=https;branch=master \
 	file://can.rules \
+    file://fb.rules \
```
agherzan, line 8: Small request: please maintain the indentation.

agherzan (changes requested):

agherzan:
Also, could you please look into the branch conflict with master?

agherzan (changes requested):
Thanks again for this. Could you tweak your commit message to be in accordance with our contribution guidelines? You can find them here (https://meta-raspberrypi.readthedocs.io/en/latest/contributing.html), and you can also see examples in the history. In this specific case, there si no SoB footer.

agherzan:
Another small comment: you are changing the "udev-rules-rpi" recipe, so I would use that in the subject line.

Kiwi4Inovex:
Took me awhile sorry

agherzan (approved):
Thanks!

PR #1292, opened 2024-03-02
picamera-libs: removed unused libraries from python3-picamera

carlonluca:
Apparently the recipe for python3 picamera is installing some libraries from /opt/vc/lib that are not necessary to the picamera module. The gles2 library, in particular, overwrites a symlink to another version of the gles2 library. The result is that two different gles2 libs are placed in /usr/lib.

This commit removes a few unneeded libraries, leaving libmmal and its dependencies, that are instead required.

### DIFF
--- a/recipes-multimedia/picamera-libs/picamera-libs.bb
+++ b/recipes-multimedia/picamera-libs/picamera-libs.bb
@@ -11,6 +11,7 @@ S = "${RPIFW_S}"
 do_install(){
     install -m 0755 -d ${D}${libdir}
     install -m 0755 ${S}/opt/vc/lib/*.so ${D}${libdir}
+    rm -f ${D}${libdir}/libGLES* ${D}${libdir}/libEGL* ${D}${libdir}/libWFC.so ${D}${libdir}/libOpenVG.so
 }
 
 FILES:${PN} = "${libdir}"


agherzan (changes requested):
This looks good. Could you tweak your commit subject to state the recipe name? Checkout https://meta-raspberrypi.readthedocs.io/en/latest/contributing.html for more info.

kraj:
> Apparently the recipe for python3 picamera is installing some libraries from /opt/vc/lib that are not necessary to the picamera module. The gles2 library, in particular, overwrites a symlink to another version of the gles2 library. The result is that two different gles2 libs are placed in /usr/lib.
> 
> This commit removes a few unneeded libraries, leaving libmmal and its dependencies, that are instead required.

Does py3-camera work with mesa provided GL libraries ?

carlonluca:
The concept behind this is that the module does not use GL libs at all. I looked into the code and could only find libmmal.so and libbcm_host.so. GL does not seem to be a dependency.

kraj:
> The concept behind this is that the module does not use GL libs at all. I looked into the code and could only find libmmal.so and libbcm_host.so. GL does not seem to be a dependency.

OK, in that case, we should not install those either. If python pieces are abstacted enough that it does not demand MMAL
interface then it is good.

carlonluca:
Not sure if I'm understanding correctly what you are saying @kraj. libmmal and libbcm_host appear to be needed, so those, and their deps, should not be removed:

https://github.com/waveform80/picamera/blob/master/picamera/mmal.py#L49
https://github.com/waveform80/picamera/blob/master/picamera/bcm_host.py#L47

I'm not a picamera dev though.

agherzan:
@kraj I think this specific removal makes sense. @carlonluca, have you managed to ask upstream why they include the GL libraries?

carlonluca:
I'm sorry, the situation is not completely clear to me. Can you provide more details? By "upstream" here, do you mean the authors of that python library (I guess this is the official repo: https://github.com/waveform80/picamera)? Where did they include the GL libraries?

agherzan:
Scratch that last comment of mine. This is from the rpi firmware repo.

recipes-multimedia/picamera-libs/picamera-libs.bb
```diff
@@ -9,6 +9,7 @@ include recipes-bsp/common/raspberrypi-firmware.inc
 S = "${WORKDIR}/raspberrypi-firmware-1.${PV}"
 
 do_install(){
+    rm -f ${S}/opt/vc/lib/libGLES* ${S}/opt/vc/lib/libEGL* ${S}/opt/vc/lib/libWFC.so ${S}/opt/vc/lib/libOpenVG.so
```
agherzan, line 12: Shouldn't this `rm` come after the `install` command bellow?

carlonluca: I'm not experienced with bitbake, but I thought this was pretty much the same thing. My understanding is that the firmware files are extracted in ${S}/opt/vc/lib, and then all copied to the sysroot.  I thought we could copy and remove or simply remove and not copy at all.

agherzan: The install rule is the one you have in the bb file. There is nothing before (as this is not an append). So, I'm not sure what the aim is here. Have you tried your approach?

carlonluca: It seemed to solve the problem for me. The wrong libGLES library was not in the sysroot anymore.
Did you have a different result?

agherzan: Right. I missed one thing - that this removes from S. I would avoid that mutating S. Maybe clean up the install later or do an rm on D after things get installed.

carlonluca: I changed as you requested. It seems to still work for me.

agherzan:
@kraj, I see you had a comment on this. Thumbs up on the current patch?

agherzan:
@kraj small reminder on this just in case you missed it.

kraj (approved):
lgtm

agherzan (approved):

PR #1294, opened 2024-03-03
 libcamera bug fix

smzahraee:
on Scarthgap yocto version libcamera crash due to unbuild its dependencies.
so I added libcamera branch of raspberry and also add libpisp to the multimedia-layer


shr-project:
This will cause all MACHINEs to use rpi's libcamera fork as dynamic-layers/multimedia-layer/recipes-multimedia/libcamera/libcamera_0.2.0.bb doesn't even restrict COMPATIBLE_MACHINE to rpi.

I don't use camera on rpi, but it would be better to find a way for these two to co-exist nicely (even if it means applying rpi specific patches on top of upstream libcamera - with rpi override).

agherzan:

PR #1298, opened 2024-03-20
raspberrypi-firmware: Resort to github APIs for tarballs

kraj:
The firmware repo is huge ( 36G ) to clone is onerous using shallow clone tarballs require them to be pre-populated at some mirror which we don't have. Third option is to let github provide the tarball, there is concern that tarballs might change the compression etc and thus checksum would change that is something github would communicate to users in advance [1]

Therefore this is the best solution for us right now.

Also upgrade to latest while here and drop PR settings

[1] https://docs.github.com/en/repositories/working-with-files/using-files/downloading-source-code-archives#stability-of-source-code-archives

**- What I did**

**- How I did it**


shr-project:
I think the previous firmware update broke picamera-libs:
do_install: install: cannot stat 'picamera-libs/20240312/git/opt/vc/lib/*.so': No such file or directory
do_populate_lic: ERROR: QA Issue: lib32-picamera-libs: LIC_FILES_CHKSUM points to an invalid file: picamera-libs/20240312/git/opt/vc/LICENCE [license-checksum]

I don't use picamera-libs nor python3-picamera (which depends on picamera-libs), aren't they replaced by libcamera stuff?

https://github.com/agherzan/meta-raspberrypi/pull/1292 probably won't fix it as well, neither does https://github.com/agherzan/meta-raspberrypi/pull/1294

kraj:
I think picamera should be replaced with https://github.com/raspberrypi/picamera2 since it supports 64bit as well as 32bit, unlike picamera which only works with precompiled userland VC drivers and 32bit only.

kraj:
> I think the previous firmware update broke picamera-libs: do_install: install: cannot stat 'picamera-libs/20240312/git/opt/vc/lib/*.so': No such file or directory do_populate_lic: ERROR: QA Issue: lib32-picamera-libs: LIC_FILES_CHKSUM points to an invalid file: picamera-libs/20240312/git/opt/vc/LICENCE [license-checksum]
> 
> I don't use picamera-libs nor python3-picamera (which depends on picamera-libs), aren't they replaced by libcamera stuff?
> 
> #1292 probably won't fix it as well, neither does #1294

I have separated 32bit only recipe needs from rpi-bootfiles and that should fix the issue.

paolobarbolini:
I'm wandering whether `--filter=blob:none` could fix the issue, if bitbake supports passing that flag. For me it took 20 seconds to clone the repo from scratch and the entire directory is about 480 MB

kraj:
> I'm wandering whether `--filter=blob:none` could fix the issue, if bitbake supports passing that flag. For me it took 20 seconds to clone the repo from scratch and the entire directory is about 480 MB

sadly, bitbake fetcher does not accept git commands directly, it has to be modified for that.

kraj:
@agherzan we need to merge this quickly, or else everyone will be fetching ~36G of firmware repo and break the internet :)

nandra:
@agherzan can you please merge this. It's blocking me to use meta-raspberrypi master branch for test build. Thanks.

agherzan (approved):
Thanks! Merged!

PR #1310, opened 2024-04-22
fix conflicting boot.scr if UBOOT_ENV is set

asac:


asac:
please also cherry-pick to kirkstone branch if good. thx.

recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -6,7 +6,7 @@ SRC_URI:append:rpi = " \
 
 SRC_URI:append:rpi = " file://0001-rpi-always-set-fdt_addr-with-firmware-provided-FDT-address.patch"
 
-DEPENDS:append:rpi = " u-boot-default-script"
+DEPENDS:append:rpi = " ${@oe.utils.conditional('UBOOT_ENV', '', 'u-boot-default-script', '', d)}"
```
agherzan, line 9: I think this is a good idea. There are some things that I would refine: I think the core also checks for UBOOT_ENV_SUFFIX (or similar).

agherzan (changes requested):
Thanks for this! I think this makes sense. Could you reformat a bit the commit using https://meta-raspberrypi.readthedocs.io/en/latest/contributing.html?

asac:
OK @agherzan, I improved the commit message. From grepping the core code it seems most code that decides whether to install something or not is really hooked on UBOOT_ENV ... also UBOOT_ENV_SUFFIX has a default so it would not be an empty check. I suggest we land this as is and see if someone else runs into issues.

recipes-bsp/u-boot/u-boot_%.bbappend
asac: OK @agherzan, I improved the commit message. From grepping the core code it seems most code that decides whether to install something or not is really hooked on UBOOT_ENV ... also UBOOT_ENV_SUFFIX has a default so it would not be an empty check. I suggest we land this as is and see if someone else runs into issues.

kraj, line 9: I think its better to tweak `u-boot-default-scripts` recipe via a bbappend in such a distro.  This is introducing a variable that wont be used elsewhere.

kraj (changes requested):

recipes-bsp/u-boot/u-boot_%.bbappend
asac: Hey @kraj UBOOT_ENV is not a new variable invented by us. From what I read in the code it's the "official" core way of adding a boot.scr in a standardized manner ... https://github.com/openembedded/openembedded-core/blob/master/meta/classes-recipe/uboot-config.bbclass#L61 ... the u-boot_%.bbappend of meta-raspberrypi breaks that behaviour right now.

Maybe we should also check if template support for this in openembedded-core would be ok.

kraj: right I did not check that the variable is already in core layer. However, this variable is relevant for deploy tasks. the logic is still twisted. IMO this change seems more appropriate in u-boot-defaults-script recipe for meta-rpi since this recipe provides the UBOOT_ENV expected file.

agherzan:
This looks abandoned. CLosing for now, but feel free to reopen it if needed.

PR #1320, opened 2024-05-14
[scarthgap] linux-raspberrypi: Drop 5.15

shr-project:
* 6.1 version was used by default since mickledore:
  https://github.com/agherzan/meta-raspberrypi/commit/12a1187b01644cb444106f7c8f498fbf519be276
  and kirkstone for rpi5 MACHINEs since:
  https://github.com/agherzan/meta-raspberrypi/commit/9dc6673d41044f1174551120ce63501421dbcd85

* scarthgap uses 6.6 by default since:
  https://github.com/agherzan/meta-raspberrypi/commit/1cf3dd5e5ef86c5196c8544853857ab1561c748b

* nobody is probably using 5.15 with newer release and there are issues with "iw reg" since
  wireless-regdb upgrade in oe-core as well as shown in:
  https://github.com/agherzan/meta-raspberrypi/pull/1317
  so better to drop it from master/scarthgap together with
  rpi.scc used only for 5.15 and 0001-gcc-plugins-Reorganize-gimple-includes-for-GCC-13.patch
  used only by rpi.scc


kraj (approved):
lgtm

agherzan (approved):

PR #1326, opened 2024-05-23
recipes: Switch to using UNPACKDIR instead of WORKDIR

kraj:
UNPACKDIR is new contruct for do_unpack things in latest master we should be using that instead of WORKDIR for referencing those files


kraj:
Please note that this change would not work on scarthgap so we need to branch scarthgap before applying this patch

shr-project:
You might have missed one: https://github.com/agherzan/meta-raspberrypi/pull/1329

agherzan (approved):
Looks good! Thanks.

agherzan:
Note that I've also created "scarthgap".

PR #1349, opened 2024-06-30
raspi-utils: added new recipe

vermaete:
For now only the 'dtmerge' (including 'dtoverlay' and 'dtparam') and 'pinctrl' utilities are included.

Added the 'raspi-' prefix to make it less conflicting with other 'utils' recipes.


agherzan (approved):
Cheers! Looks good.

PR #1360, opened 2024-08-23
mesa: rename bbappend to match new recipe name from oe-core

shr-project:
Otherwise it will fail to parse.

Renamed in:
https://git.openembedded.org/openembedded-core/commit/?id=f5cfb3e23603cefb2f3f6bfe776afaedefd10808


kraj (approved):
lgtm

PR #1372, opened 2024-09-26
conf/layer.conf: Remove meta-lts-mixins

leon-anavi:
Remove layer dependency from meta-lts-mixins 'scarthgap/u-boot' because now the BSP supports styhead which brings U-Boot 2024.07.

**- What I did**

Removed layer dependency from meta-lts-mixins 'scarthgap/u-boot'. Historically, this dependency was added temporary to provide U-Boot 2024.04 instead of 2024.01 because of Raspberry Pi 5.

**- How I did it**

Modified `conf/layer.conf`


kraj (approved):
lgtm

agherzan (approved):
Thanks!

PR #1379, opened 2024-10-11
conf/layer.conf: Remove duplicated BBFILES

robertlinux:
The following line doubles BBFILES for all layers in BBFILE_COLLECTIONS: BBFILES += "${@' '.join('${LAYERDIR}/dynamic-layers/%s/recipes*/*/*.bb' % layer \
               for layer in BBFILE_COLLECTIONS.split())}"

And most of them are invalid, use BBFILES_DYNAMIC is the correct solution.

**- What I did**

**- How I did it**


agherzan (approved):
Looks good to me.

PR #1392, opened 2024-12-06
linux-raspberrypi-6.6: Upgrade to 6.6.63

kraj:


agherzan (changes requested):
It seems like some machines fail: "| make[3]: *** No rule to make target 'arch/arm/boot/dts/broadcom/bcm2712-rpi-5-b.dtb'.  Stop."

kraj:
> It seems like some machines fail: "| make[3]: *** No rule to make target 'arch/arm/boot/dts/broadcom/bcm2712-rpi-5-b.dtb'. Stop."

yeah, its because rpi-5 related DTS files are now moved to under `arm64` folder so we can not include it
by default on all rpi machines. Uploaded a v2 with fix.

agherzan (approved):
Nice. Thanks for this!

PR #1399, opened 2024-12-29
Some Kas stuff

vermaete:
* Added, but in comment, the synaptics-killswitch license.  Because it's easier to have in in the KAS file.

* The 'debug-tweaks' image_feature is removed in Yocto.
  I replaced it with was under it.  But this is (and was) insecure.
  **But I would be in favor of having this replaced by a more secure settings**.
  It could be in comment as a warning to the user.

br


agherzan (approved):
Looks good. Thanks for this.

PR #1417, opened 2025-02-14
rpi-base: Add missing broadcom/ prefix to find DTB files

acostach:
This is a backport of 6a7aac79ce1dc5331ec820796592797e3a66c41d from the master branch. 
It allows for building device-trees in kernel 6.6, using kirkstone.


acostach:
Hi @kraj , does this look ok?

kraj (approved):
lgtm

acostach:
looks like this won't work for kernel 5.15...

kraj:
> looks like this won't work for kernel 5.15...

thats right.

agherzan (changes requested):
We need to fix the builds.

PR #1423, opened 2025-02-21
udev-rules-rpi: Delete fb.rules

kraj:
OE-core has added fb.rules to psplash via [1] so avoid adding it here

This effectively reverts commit e9e5efa750490cd0dac853eb65640d0cb337b03e

[1] https://git.yoctoproject.org/poky/commit/?id=af38235305fc2ac59a82f0413a1fe4cc5161ff86

Cc: Bastian Wanner <bastian.wanner@inovex.de>


kraj:
@Kiwi4Inovex FYI

Issue #1306, opened 2024-04-11
Kernel Uncompress Failure on Raspberry Pi 4 in Scarthgap Yocto Release


mdrodrigo:
**Description:**

Issue:
Encountering a error during kernel uncompression on the Raspberry Pi 4 when utilizing the Scarthgap Yocto Release and meta-raspberrypi layer on the master branch. This issue manifests specifically with kernel version 6.6 on the raspberrypi4 machine. Upon booting, following the u-boot initialization, the kernel uncompression process triggers an error, causing the board to enter a reboot loop.

Error Log from U-Boot:

```
U-Boot 2024.01 (Jan 08 2024 - 15:37:48 +0000)

DRAM:  628 MiB (effective 3.6 GiB)
RPI 4 Model B (0xc03111)
Core:  212 devices, 16 uclasses, devicetree: board
MMC:   mmcnr@7e300000: 1, mmc@7e340000: 0
Loading Environment from FAT... OK
In:    serial,usbkbd
Out:   serial,vidconsole
Err:   serial,vidconsole
Net:   eth0: ethernet@7d580000
PCIe BRCM: link up, 5.0 Gbps x1 (SSC)
PCI: Failed autoconfig bar 10
starting USB...
Bus xhci_pci: dm_pci_bus_to_phys: invalid physical address
xhci-pci init cannot map PCI mem bar
probe failed, error -5
No working controllers found
Hit any key to stop autoboot:  2  1  0 
Card did not respond to voltage select! : -110
** Booting bootflow 'mmc@7e340000.bootdev.part_1' with script
Working FDT set to 273e2400
8520088 bytes read in 370 ms (22 MiB/s)
## Booting kernel from Legacy Image at 00080000 ...
   Image Name:   Linux-6.6.22-v7l
   Image Type:   ARM Linux Kernel Image (uncompressed)
   Data Size:    8520024 Bytes = 8.1 MiB
   Load Address: 00008000
   Entry Point:  00008000
   Verifying Checksum ... OK
## Flattened Device Tree blob at 273e2400
   Booting using the fdt blob at 0x273e2400
Working FDT set to 273e2400
   Loading Kernel Image
uncompressed: uncompress error -28
Must RESET board to recover
resetting ...
B/
```

**Observations:**
Curiously, this issue does not persist when utilizing the raspberrypi4-64 variant. In this case, the kernel operates without encountering any issues.

**Steps to Reproduce:**

[...]

PR #1432, opened 2025-03-06
u-boot_%.bbappend: Increase CONFIG_SYS_BOOTM_LEN

leon-anavi:
**- What I did**

This patch fixes https://github.com/agherzan/meta-raspberrypi/issues/1306, aka the following issue when booting an image with U-Boot:

```
   Image Name:   Linux-6.6.63-v7l
   Image Type:   ARM Linux Kernel Image (uncompressed)
   Data Size:    8559752 Bytes = 8.2 MiB
   Load Address: 00008000
   Entry Point:  00008000
   Verifying Checksum ... OK
   Booting using the fdt blob at 0x2eff2000
Working FDT set to 2eff2000
   Loading Kernel Image to 8000
Image too large: increase CONFIG_SYS_BOOTM_LEN
Must RESET board to recover
```

**- How I did it**

Increased `CONFIG_SYS_BOOTM_LEN` for machine `raspberrypi4` (32-bit) from 8MB to 16MB because the kernel size has increased.


kraj (approved):
lgtm

PR #1442, opened 2025-04-01
ci: Update build container images to Ubuntu 24.04

agherzan:
This fixes the latest Python version requirement.

### DIFF
--- a/.github/workflows/docker-images/yocto-builder/Dockerfile
+++ b/.github/workflows/docker-images/yocto-builder/Dockerfile
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: MIT
 
-FROM ubuntu:20.04
+FROM ubuntu:24.04
 
 ARG DEBIAN_FRONTEND="noninteractive"
 RUN apt-get update -qq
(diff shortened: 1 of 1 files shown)


kraj:
we might need to tweak the system apparmor policy see

https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2056555

here is what worked for me on 24.04

```
   sudo tee /etc/apparmor.d/bitbake > /dev/null <<'EOF'
   abi <abi/4.0>,
   include <tunables/global>
   profile bitbake /**/bitbake/bin/bitbake flags=(unconfined) {
           userns,
   }
   EOF
```

then apply it

```
   # Reload AppArmor Profile
   sudo apparmor_parser -R /etc/apparmor.d/bitbake
```

Sometimes this may not be enough. Then do below it will compromise apparmor defaults a bit more.
```
   sudo apparmor_parser -R /etc/apparmor.d/unprivileged_userns
```

agherzan:
@kraj It doesn't look like it broke due to AppArmor. What do you think of the logs?

kraj:
> @kraj It doesn't look like it broke due to AppArmor. What do you think of the logs?

interesting, maybe your install does not enable it, but in anycase if thats not being flagged then I am ok with this patch

kraj:
> @kraj It doesn't look like it broke due to AppArmor. What do you think of the logs?

We are not mounting any persistent directory inside the docker right ? since the UID has changed it might be an issue
secondly, I usually use

umask 0022

and launch the docker with --ulimit "nofile=1024:1048576" to overcome file handle issues when building bigger components like chromium

agherzan:
Partially superseeded by https://github.com/agherzan/meta-raspberrypi/pull/1458. An upgrade to 24.04 can be deferred for now.

PR #1457, opened 2025-04-12
doc: added raspberrypi5 to the list of supported machines

vermaete:


docs/layer-contents.md
```diff
@@ -11,6 +11,7 @@
 * raspberrypi3-64 (64 bit kernel & userspace)
 * raspberrypi4
 * raspberrypi4-64 (64 bit kernel & userspace)
+* raspberrypi5
```
stu-spp, line 14: We also overload this for the cm5. So maybe:
```
raspberrypi5 (64 bit kernel & userspace & also used for the cm5)
```

stu-spp (approved):

Issue #1313, opened 2024-04-30
psplash issues with systemd


leon-anavi:
Hi,

I noticed that psplash doesn't work as expected when using it with systemd despite [the fix for race condition](98349bb367418866e008ba22ca75f4ab2984561a) with /dev/fb0:

```
Apr 30 11:04:09 raspberrypi4-64 systemd[1]: sys-devices-platform-gpu-graphics-fb0.device: Job sys-devices-platform-gpu-graphics-fb0.device/start timed out.
Apr 30 11:04:09 raspberrypi4-64 systemd[1]: Timed out waiting for device /sys/devices/platform/gpu/graphics/fb0.
Apr 30 11:04:09 raspberrypi4-64 systemd[1]: sys-devices-platform-gpu-graphics-fb0.device: Job sys-devices-platform-gpu-graphics-fb0.device/start failed with result 'timeout'.
```

**Steps to reproduce the issue:**

1. Include psplash: `IMAGE_INSTALL:append = " psplash"`
2. Use systemd: `INIT_MANAGER = "systemd"`

**Describe the results you received:**

No splash screen. Systemd service psplash-start dependency failed:

```
Apr 30 11:04:09 raspberrypi4-64 systemd[1]: sys-devices-platform-gpu-graphics-fb0.device: Job sys-devices-platform-gpu-graphics-fb0.device/start timed out.
Apr 30 11:04:09 raspberrypi4-64 systemd[1]: Timed out waiting for device /sys/devices/platform/gpu/graphics/fb0.
Apr 30 11:04:09 raspberrypi4-64 systemd[1]: sys-devices-platform-gpu-graphics-fb0.device: Job sys-devices-platform-gpu-graphics-fb0.device/start failed with result 'timeout'.
```

**Describe the results you expected:**

`psplash-start.service` to be OK and to display the splash screen while booting the system

**Additional information you deem important (e.g. issue happens only occasionally):**

**Additional details (revisions used, host distro, etc.):**

Tested with core-image-base from branches master and kirkstone on Raspberry Pi 4 (machine `raspberrypi4-64`).

farias-automac:
Same

BuriKizilkaya:
Have you found a solution?

gitgonewithwind:
@BuriKizilkaya 
please try my solution:
https://github.com/agherzan/meta-raspberrypi/issues/1330#issuecomment-2142679662

BuriKizilkaya:
[...]

PR #1474, opened 2025-05-20
Revert "psplash: Make psplash wait for the framebuffer to be ready"

Renaud-Barrau:
Hi,
Due to a confused commit history between poky (https://github.com/yoctoproject/poky) and meta-raspberrypi (https://github.com/agherzan/meta-raspberrypi) both on branch **scarthgap**,  it seems that's there is still an issue between psplash and systemd, which is explained in this issue #1313 .

I propose in my fix to revert commit  98349bb367418866e008ba22ca75f4ab2984561a.

**- What I did**
My project aim to build raspberry pi 3B+ Yocto image using Kas. I'm fetching _Poky_ on branch **scarthgap**, as well as _meta-raspberrypi_ on branch **scarthgap**
When starting the rasp, the same error as #1313  shows up.
As mentioned in this issue, it seems to be fixed in #1268, but I can't find any of it in scarthgap

I believe this way to fix the issue might not be the best one but it's the most simple and efficient I found, in my particular configuration detailled above.

I also want to apologize in advance for obvious mistakes I may have done, I 'm fairely new to the meta-raspberrypi and Yocto environnement :)


kraj:
Can you also backport https://github.com/agherzan/meta-raspberrypi/commit/e9e5efa750490cd0dac853eb65640d0cb337b03e to scarthgap with this same PR

Renaud-Barrau:
I believe the error in CI is due to the transition from ${WORKDIR} to ${UNPACKDIR} in 6e0bae7, which is supported from stylehead version, but not in scarthgap ? 
https://docs.yoctoproject.org/next/migration-guides/migration-5.1.html

kraj:
Can you rebase once again please ?

agherzan:
@kraj This seems to fail on the cherry-picked udev rules commit.

kraj:
@Renaud-Barrau maybe its no more a simple revert+backport anymore can you rework it

Renaud-Barrau:
I have done some tests today and in my case, the revert alone works well on my rpi3. I can't find any differencies in the behaviour of the system with or without the udev rules commit e9e5efa. But since @kraj mention it, I believe this is important so I fixed the UNPACKDIR -> WORKDIR issue in it (which exists because of the scarthgap branch). Now the CI should run fine :smile:

Renaud-Barrau:
Done

a-elkaim:
Hi @Renaud-Barrau i've been using your branch (that is supposed to contain the fixes?) but I still don't get a splash using core-image-weston as an image.

Moreover, I have a very weird behavior:

- When I use `agherzan/meta-raspberrypi` scarthap or when I use your scarthgap branch with the updated `psplash_%.bbappend`
   - No splash is displayed at all
   - The psplash binary (`/usr/bin/pslash`) **is NOT present** on the filesystem
- When I remove meta-raspberrypi's `psplash_%.bbappend` (rename it to psplash_%.bbappend.ignore)
    - Poky's default splash is correctly displayed
    - The psplash binary (`/usr/bin/pslash`) is present on the filesystem

So I added the following to your version of psplash_%.bbappend (your fixes are needed but are not enough, as I said, for some reason the psplash binary `/usr/bin/pslash` is not present on the filesystem):

```
do_install:append() {
    install -m 0755 ${B}/psplash ${D}${bindir}/psplash
}
```

And now I correctly have the splash being displayed at boot time.

I'm really not sure what is going on here.

Renaud-Barrau:
Hello @azsde, I have done some research about your issue.

First, I can say with certainty that your issue is not related to this PR's commits.

Indeed, as you said, `/usr/bin/pslash` is not installed in the target rootfs, but this is the "normal" behaviour of the main psplash recipe, as you can see in the `psplash_git.bb` of openembedded-core layer line 119 (https://github.com/openembedded/openembedded-core/blob/scarthgap/meta/recipes-core/psplash/psplash_git.bb)

````
	install -d ${D}${bindir}
	for i in ${SPLASH_INSTALL} ; do
		install -m 0755 $i ${D}${bindir}/$i
	done
	rm -f ${D}${bindir}/psplash
````

As far as I understand, this is done to "clean" the psplash binaries, as only the ones that are suffixed (`/usr/bin/psplash-raspberrypi` for example) are installed.

BUT, I found a user with a similar issue in the mail archive, saying that the transition between Kirkstone and Scarthgap broke the old outsuffix ! (I cannot understand why :sweat_smile: ) : https://www.mail-archive.com/openembedded-core@lists.openembedded.org/msg193752.html

So one solution should be to set the outsuffix to "default" in your custom `psplash_git.bbappend` with the variable SPLASH_IMAGES, at the same time as you change the psplash image : 
````
SPLASH_IMAGES = "file://my-custom-psplash-image.png;outsuffix=default"
````
Another solution could be to keep your 
````
do_install:append() {
    install -m 0755 ${B}/psplash ${D}${bindir}/psplash
}
````
 
in your custom layer `psplash_git.bbappend`. (which is also the solution I use in my project btw)

Theses workaround should work well, but this means there might be another issue with outsuffix in `scarthgap`.

Maybe @agherzan and @kraj could enlighten us ? :smile:

PR #1491, opened 2025-07-07
Whinlatter update ( consolidated )

kraj:

PR #1492, opened 2025-07-08
picamera-libs additional fix for S/UNPACKDIR

shr-project:
https://github.com/agherzan/meta-raspberrypi/pull/1491 + one more fix for picamera-libs


shr-project:
Thanks

PR #1494, opened 2025-07-16
docker-build: use --no-cache

shr-project:

* use buildx to avoid: DEPRECATED: The legacy builder is deprecated and will be removed in a future release. Install the buildx component to build images with BuildKit: https://docs.docker.com/go/buildx/


kraj (approved):
LGTM

shr-project:
Added to https://github.com/agherzan/meta-raspberrypi/pull/1491 to fix the docker build as well as whinlatter compatibility at the same CI run.

Will submit kirkstone, scarthgap, walnascar PRs soon (jansa/kirkstone-docker, .. branches are already created)

PR #1504, opened 2025-08-07
[scarthgap] mesa, wayland-protocols: use separate recipe instead of bbappend

shr-project:
Fix
https://github.com/agherzan/meta-raspberrypi/pull/1484

Avoid llvm runtime dependency to avoid issues in builds with meta-clang:

ERROR: Multiple .bb files are due to be built which each provide llvm:
  meta-clang/recipes-devtools/clang/clang_git.bb
  oe-core/meta/recipes-devtools/llvm/llvm_18.1.6.bb
A list of tasks depending on these providers is shown and may help explain where the dependency comes from. meta-clang/recipes-devtools/clang/clang_git.bb has unique dependees:
  oe-core/meta/recipes-graphics/mesa/mesa_24.0.7.bb:do_prepare_recipe_sysroot
  oe-core/meta/recipes-graphics/mesa/mesa_24.0.7.bb:do_package
  meta-atlas/meta-atlas/recipes-core/images/atlas-image-video.bb:do_populate_lic_deploy
  oe-core/meta/recipes-graphics/mesa/mesa_24.0.7.bb:do_collect_spdx_deps
  meta-atlas/meta-atlas/recipes-core/images/atlas-image-video.bb:do_populate_interfaces_deploy
oe-core/meta/recipes-devtools/llvm/llvm_18.1.6.bb has unique dependees:
  oe-core/meta/recipes-graphics/mesa/mesa_24.0.7.bb:do_package_qa
  oe-core/meta/recipes-graphics/mesa/mesa_24.0.7.bb:do_create_runtime_spdx
It could be that one recipe provides something the other doesn't and should. The following provider and runtime provider differences may be helpful. meta-clang/recipes-devtools/clang/clang_git.bb has unique provides:
  clang
meta-clang/recipes-devtools/clang/clang_git.bb has unique rprovides:
  clang-dbg
  clang-libclang-cpp
  liblldb
  llvm-linker-tools
  ^clang-locale-.*
  clang-libllvm
  lldb
  clang-format
  clang
  clang-dev
  clang-locale
  clang-lldb-python
  libclang
  clang-src
  clang-doc
  clang-staticdev
  lldb-server
  clang-tidy
  clang-tools
oe-core/meta/recipes-devtools/llvm/llvm_18.1.6.bb has unique provides:

oe-core/meta/recipes-devtools/llvm/llvm_18.1.6.bb has unique rprovides:
  llvm
  llvm-dbg
  llvm-libllvm
  llvm-src
  llvm-dev
[...]

PR #1484, opened 2025-06-27
[Scarthgap] Upgrade to Mesa 25.1.6 and wayland-protocols 1.45

leon-anavi:
**- What I did**

Added Mesa 25.1.3 and wayland-protocols 1.45 to branch Scarthgap. Tested it on Raspberry Pi 5 with the official Raspberry Pi Touch Display 2.

**- How I did it**

* Cherry-picked @BassemMohsen patch from: https://github.com/agherzan/meta-raspberrypi/pull/1395
* Updated wayland-protocols to version 1.45 because new Mesa version requires wayland-protocols 1.41 or newer
* Updated Mesa from 24.3.1 -> 25.1.3

Mesa version 25.1.3 contains the following [fix](https://docs.mesa3d.org/relnotes/25.1.0.html) related to #1483:

- v3dv: vkcube-wayland crashes on raspberry pi 5 kernel 6.12 and latest mesa

PR #1506, opened 2025-08-10
rpi-config: fix VC4DTBO names for rpi4 and rpi5

10ne1:
**- What I did**

Had problems getting rpi5 v3d acceleration with glamor, xserver, wayland, weston, sway etc. Full of crashes, non-deterministic performance, freezes, irregular frame updates and so on. Software rendering, on the other hand worked without issues (slower as expected, however it worked consistently).

**- How I did it**

After some hours spent debugging, going lower and lower level from xserver/wayland to mesa, the kernel, finally I arrived at the device trees and noticed rpi5 and rpi4 are supposed to use specialised VC4DTBO. It's a miracle the generic legacy DTBO worked at all.

### DIFF
--- a/recipes-graphics/mesa/mesa/0001-dont-build-clover-frontend.patch
+++ b/recipes-graphics/mesa/mesa/0001-dont-build-clover-frontend.patch
@@ -1,0 +1,29 @@
+From: Markus Volk <f_l_k@t-online.de>
+Date: Sun, 19 Mai 2025 15:34:46 +0100
+Subject: [PATCH] dont build clover frontend
+
+The clover frontend is deprecated and is always built with opencl, even if
+using rusticl. Additionally it adds a reproducibility issue.
+
+Upstream-Status: Inappropriate [oe-specific]
+Signed-off-by: Markus Volk <f_l_k@t-online.de>
+
+--- a/src/gallium/meson.build	2025-05-07 18:35:10.000000000 +0200
++++ b/src/gallium/meson.build	2025-05-18 17:05:23.677694272 +0200
+@@ -195,15 +195,11 @@
+ else
+   driver_d3d12 = declare_dependency()
+ endif
+-if with_gallium_clover or with_tests
++if with_tests
+   # At the moment, clover and gallium/tests are the only two consumers
+   # for pipe-loader
+   subdir('targets/pipe-loader')
+ endif
+-if with_gallium_clover
+-  subdir('frontends/clover')
+-  subdir('targets/opencl')
+-endif
+ if with_gallium_rusticl
+   subdir('frontends/rusticl')
+   subdir('targets/rusticl')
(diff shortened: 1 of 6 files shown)


shr-project:
Is this new with kernel 6.12 or when did the DTBO name change?

shr-project:
> Is this new with kernel 6.12 or when did the DTBO name change?

Based on 
https://forums.raspberrypi.com/viewtopic.php?t=365534
https://github.com/raspberrypi/linux/commit/1196bf1a7736ff0ab79f5012fa84082e298031a7#diff-d2043d647c473acf49527946234854f35dbba27b2f8a341cd88a2da07dd6f728R434

And it's still there in 6.12:
https://github.com/raspberrypi/linux/blob/rpi-6.12.y/arch/arm/boot/dts/overlays/overlay_map.dts#L510

the vc4-kms-v3d value is still supposed to work, right? Are you sure it doesn't?

10ne1:
Hello, I have been testing both v6.6 and v6.12 on rpi5 and 3d acceleration is broken with `dtoverlay=vc4-kms-v3d` in the following way:

All drivers successfully probe, glxinfo & such report that v3d acceleration is correctly enabled, the /dev/dri/cards and render nodes get "correctly" picked up, all processes and desktop environments start in the background and function, yet when it's time to render... just one or two frames get actually rendered, then the system randomly freezes, requiring a hard power cycle.

`dtoverlay=vc4-kms-v3d-pi5` makes it work as expected. I haven't dug deeper than this, took me quite a while to figure out. Maybe I can find some more time to investigate in the coming days or weekend.

shr-project:
Can you please check that overlay_map is installed in your system? It's included in RPI_KERNEL_DEVICETREE_OVERLAYS so unless you modify KERNEL_DEVICETREE somewhere it should be there.

Of course there might be some bug somewhere, I believe you and your testing, just trying to understand it better myself (and decide if we might need this for other branches as well). I will try to check on rpi5 as well (because it doesn't crash for us) and on rpi4 (where we used vc4-fkms-v3d vc4-kms-v3d until recently both also without issues).

10ne1:
@agherzan @shr-project 

Turns out the VC4DTBO was a red herring. Indeed the map file and all dtbo are present and work as expected.

I was thrown off due to an `Option "AccelMethod" "none"` in my configs which used software rendering.

The real root cause is the old Mesa 24.0.7 from the `walnascar` branch. I noticed very recently mesa 25.1.6 landed in the `scathergap` branch, so I cherry-picked that and it works. Apparently there are some v3d fixes in newer mesa which are required.

Can you please cherry-pick 3afc9728b1f4ba0f5be1af34883d6582966133a1 to `walnascar` ?

10ne1:
Closing, please just cherry-pick https://github.com/agherzan/meta-raspberrypi/commit/3afc9728b1f4ba0f5be1af34883d6582966133a1 to the walnascar branch :)

agherzan:
@shr-project I see that you handled the commit mentioned here. Is that specific to scarthgap?

shr-project:
I've created that commit for scarthgap just to avoid breakage caused by https://github.com/agherzan/meta-raspberrypi/pull/1484
and in my builds I'm still using older mesa.

So if it does fix some runtime issues, then I guess https://github.com/agherzan/meta-raspberrypi/pull/1484 could be applied in walnascar as well together with https://github.com/agherzan/meta-raspberrypi/pull/1504

agherzan:
@shr-project Cheers for clarifying. @10ne1 would you be up for creating and testing a PR with the commits mentioned above?

10ne1:
Yes @agherzan I can create a PR since I already have them applied locally for testing.

Will likely have to wait after OSS EU and the Git mini-summit because I have to prepare some talks for those next week.

PR #1538, opened 2025-11-05
mesa: rename bbappend

sapiippo:
Rename the bbappend to apply to the newly added mesa recipe.


agherzan:
@sapiippo could you link this to what change generated this incompatibility? Also, I suspect this affects master, too. Could you create a PR for it as well?

sapiippo:
This is needed after commit 5c540d5 that added mesa recipe with version in the filename (mesa_25.1.6.bb). That change is only available in walnascar. Since oe-core/master provides mesa recipe without version number (mesa.bb), this change is also needed only in walnascar.

PR #1547, opened 2025-11-20
Upgrade 6.12/6.6/6.1 kernels to latest

kraj:


OldManYellsAtCloud:
Is that weston crash solved with this kernel? I started to bisect it (slowly), but if it is a non-issue, that would make it easier.

kraj:
> Is that weston crash solved with this kernel? I started to bisect it (slowly), but if it is a non-issue, that would make it easier.

Please try these kernels and verify in scarthgap in particular.

OldManYellsAtCloud:
I see the crash with both 6.6.78 and 6.12.58 (haven't tried the 6.1 one). Tested with rpi4-64 @ Scarthgap.

Will try to finish the bisecting to see if it shows anything

PR #1556, opened 2025-12-03
rpi-base: Add missing hat_map.dtb

flo-at:
hat_map.dtb is required by the firmware to match HAT UUIDs from EEPROM data to their corresponding device tree overlays during boot. Without this file, HATs may fail to initialize properly even when their overlay files are present.

hat_map.dtb is generated and treated the same way overlay_map.dtb is. On the boot partition, it goes into the overlays/ directory, hence the special treatment.

**- What I did**
I changed rpi-base.inc so that hat_map.dtb is deployed and installed into the boot partition at the correct path.

**- How I did it**
It behaves just as overlay_map.dtb so I copied what was already in place to correctly handle that file.
The changes are minimal.

### DIFF
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -16,6 +16,7 @@ XSERVER = " \
 
 RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
     overlays/overlay_map.dtb \
+    overlays/hat_map.dtb \
     overlays/at86rf233.dtbo \
     overlays/disable-bt.dtbo \
     overlays/disable-bt-pi5.dtbo \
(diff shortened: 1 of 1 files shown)


agherzan:
Oh great! Yet another map file :)

Do you have some documentation on it? Also, could you update your git commit log format as per the contribution guidelines? 

Thanks for this!

flo-at:
Documentation for this is lacking unfortunately. It came with [this commit](https://github.com/raspberrypi/linux/commit/3a44d1cb497e2c28bdcbda21369111287be953db) (also [here](https://github.com/raspberrypi/linux/blob/45d3dba58f9996d4b475c952223841ea5da905ab/arch/arm/boot/dts/overlays/Makefile#L3)).
There's nothing mentioned in the [RPI config docs](https://www.raspberrypi.com/documentation/computers/configuration.html) yet. I only found out about this when I compared a RPI OS setup to my Yocto based one that wouldn't recognize the HAT without an explicit `dtoverlay=` line in `config.txt`, while the same thing worked on RPI OS.

I'll check/update the commit log - sorry about that.

flo-at:
I created an issue in the RPI documentation repo: https://github.com/raspberrypi/documentation/issues/4264

agherzan (approved):
Fair enough. Thanks for this!

flo-at:
The failed check looks like it timed out.

PR #1566, opened 2026-01-09
Fix compilation of rpidistro-VLC on Whinlatter

gizahNL:
**- What I did**
Fix compilation of VLC on Whinlatter
**- How I did it**
Various backports (i.e.: taglib), casts, etc. to fix compilation on Whinlatter against rpidistro-ffmpeg (upstream ffmpeg changes not backported)


kraj (approved):
LGTM

agherzan (changes requested):
Could I ask you to keep the same git log format as per the contribution guidelines? Just check the history. Thanks for fixing these.

PR #1579, opened 2026-02-11
rpi-eeprom: enable native build

embetrix:
enable native build for _rpi-eeprom_ utilities which are required 
for i.e secure boot signing or other host operations


agherzan (commented):
Thanks for your contribution. What's the reason for dropping the compatible machines? And the other append Vs +=.

PR #1592, opened 2026-04-22
fscryptctl: add recipe for enabling File-based Encryption (FBE)

RajdeepBora14:
fscryptctl: add recipe for enabling File-based Encryption (FBE)

- Kernel configs for fscrypt enabled
- Added recipe changes for userpsace utility tool fscryptctl
- Ported OP-TEE v4.4.0 for key management using Trusted Keys TA
- Added TA-side and CA-side patches for successful filesystem encryption


OldManYellsAtCloud:
A couple of random thoughts:

1. Could this be broken up into smaller, more reviewable commits?
2. If this gets into Kirkstone (which has only days left before it goes EOL), it should also get into master also, or it will get lost forever.
3. It seems to have some opinionated looking things, where the intention isn't clear - not sure if intentional or accidentally committed. If intentional, they could use some explanation in the commit message. E.g. the myconfig.cfg for u-boot, or force-enabling uart in rpi-config_git.bb

RajdeepBora14:
@OldManYellsAtCloud 

The changes are just a part of a mini project I have been owrking on myself. They are not meant for integration, as I don't intend to merge the changes to the `kirkstone` branch.
Thank you!

the-gabe:
This doesn't actually hold up any of the security properties of OP-TEE though, since there is no ARM TrustZone support.

PR #1593, opened 2026-04-27
Respberrypi jchang

JianpengChang:
Small bug fixes found during integration:

- sdcard_image-rpi: guard sysv sysctl block with if/fi to avoid orphan sysctl.conf on systemd images
- layer.conf: remove duplicate parselogs_rpi test suite entries
- rpidistro-ffmpeg: add missing --disable-libv4l2 flag in PACKAGECONFIG
- rpi-config: use if/elif for mutually exclusive DWC2 modes


kraj:
Please reconsile this with #1594 as I am inclining towards accepting that change before this one.

JianpengChang:
Deleted layer.conf: remove duplicate parselogs_rpi test suite entries. Use the fix in #1594

PR #1594, opened 2026-04-27
Runqemu for raspi1ap and raspi0, and various clean-ups

jonmason:
Add the relevant entries to conf/machine/raspberrypi0.conf and conf/machine/raspberrypi.conf to get runqemu fully running.  Workaround were needed for USB issues and PM.  Attempts were made to get raspberrypi3-64 and raspberrypi4-64, but they do not fully boot to prompt.  Patches were left in so that the work can be used by a future developer.  

Broken parselogs stuff was removed.

Finally, kas code was cleaned up and broken into multiple machine specific files.  This allows for a user to simply specify the machine (and not hand edit a file to specify which machine and features).  In doing all of this, it is now possible to run CI on this setup with a simple:
kas build kas/raspberrypi.yml:kas/testimage.yml


conf/layer.conf
OldManYellsAtCloud, line 1: Just a small note: this file change will clash with PR #1593 - which I think also has a good chance to be accepted (similar to this). It looks like a fly-by improvement in both PRs, though this one seems to be a bit more complete (due to the change in rpi-default-settings.inc also) - ping @JianpengChang

jonmason: I think relevant part of PR 1593 is incomplete, as the parselogs_rpi.py is not removed.  That file is no longer relevant, as OE Core changes 2+ years ago, and parselogs now uses txt files to indicate ignores.

JianpengChang: I didn't notice the oe-core commit. My changes do have defects — please ignore the related parts.

kraj:
@jonmason is this ready ? I do see there are open questions.

conf/layer.conf
jonmason: On June 8, PR #1593 removed the commit that handles the parselogs issue.  So, given that it is no longer a potential issue.  I will resolve this conversation.

jonmason:
> @jonmason is this ready ? I do see there are open questions.

I resolved the conversation outstanding.  That issue with the other PR had been resolved for some time, but I didn't know it was on me to do anything.  Mailing lists are so much easier.

kas/raspberrypi3.yml
```diff
@@ -0,0 +1,14 @@
...
+header:
+  version: 14
+  includes:
+    - kas/base.yml
+
+local_conf_header:
+  license: |
+    # Uncomment next line to allow the license
+    # See: linux-firmware-rpidistro in docs/ipcompliance.md
+    LICENSE_FLAGS_ACCEPTED = "synaptics-killswitch"
```
kraj: This was commented out but now its enabled after rework. while the comment above it still says "Uncomment next line". docs/ipcompliance.md is explicit that this is a deliberate gate: "the build system will only allow this recipe to be built if the user acknowledges this risk". The layer's own shipped configs shouldn't perform that acknowledgment for the user,

kas/testimage.yml
```diff
@@ -0,0 +1,28 @@
...
+    # Don't testimage initramfs
+    TESTIMAGE_AUTO:pn-core-image-initramfs-boot = "0"
+  kvm: |
+    QEMU_USE_KVM = ""
+  slirp: |
+    TEST_RUNQEMUPARAMS = "slirp"
+  sshd: |
+    IMAGE_FEATURES += "ssh-server-dropbear"
+  sshkeys: |
+    CORE_IMAGE_EXTRA_INSTALL += "ssh-pregen-hostkeys"
```
kraj: this will not work for non-qemu targets. ssh-pregen-hostkeys in oe-core is COMPATIBLE_MACHINE = "^qemu.*$"

recipes-core/init-ifupdown/files/interfaces
```diff
@@ -0,0 +1,31 @@
+# /etc/network/interfaces -- configuration file for ifup(8), ifdown(8)
```
kraj: this file seems to be qemu specific but applied to all machines

kraj:
both rpi3-64 / rpi4-64 set `SERIAL_CONSOLES ?= "115200;ttyS0"`
correct for real Hardware where PL011 is wired to Bluetooth and the console is the AUX mini-UART. runqemu appends console= from SERIAL_CONSOLES and ends up with `console=ttyS0` qemuboot's default QB_SERIAL_OPT = "-serial mon:stdio -serial null" gives stdio to serial_hd(0), which in QEMU's bcm2835/bcm2711 peripherals is the PL011 (ttyAMA0) the AUX mini-UART is serial_hd(1) So your earlycon=pl011... prints on stdio 

also conf/machine/raspberrypi.conf is included by raspberrypi-cm.conf (CM1) as well as raspberrypi0.conf. So MACHINE=raspberrypi-cm now silently inherits -machine raspi1ap, QB_DTB = "bcm2708-rpi-b-plus.dtb" and IMAGE_CLASSES += "qemuboot" — wrong for a compute module

jonmason:
For the Raspberry Pi 3 and 4, it's not a case of misconfigured serial.  They are both hanging in different places, after the switch over from earlycon to serial.  I've tried many different ways and multiple attempts to get it working.  I thought it best to leave it in there in case someone else with more patience can figure it out, but I'll remove it if you don't want something fully functional in there.

I'm not sure how to best handle the raspberrypi.conf being included in raspberrypi-cm.conf.  I can move the common parts into the individual files and remove the include of raspberrypi.conf.  Or, I can drop the qemu emulation of raspberrypi (raspi1ap) completely, and only have raspi0 emulated.

FYI, the Rasperry Pi machines that can be emulated are listed at
https://www.qemu.org/docs/master/system/arm/raspi.html

kraj:
> For the Raspberry Pi 3 and 4, it's not a case of misconfigured serial. They are both hanging in different places, after the switch over from earlycon to serial. I've tried many different ways and multiple attempts to get it working. I thought it best to leave it in there in case someone else with more patience can figure it out, but I'll remove it if you don't want something fully functional in there.

Perhaps the right thing to do. You can leave a note if someone wants it.

> 
> I'm not sure how to best handle the raspberrypi.conf being included in raspberrypi-cm.conf. I can move the common parts into the individual files and remove the include of raspberrypi.conf. Or, I can drop the qemu emulation of raspberrypi (raspi1ap) completely, and only have raspi0 emulated.

Like raspberrypi-cm3.conf, maybe make a copy.

> 
> FYI, the Rasperry Pi machines that can be emulated are listed at https://www.qemu.org/docs/master/system/arm/raspi.html

jonmason:
> > For the Raspberry Pi 3 and 4, it's not a case of misconfigured serial. They are both hanging in different places, after the switch over from earlycon to serial. I've tried many different ways and multiple attempts to get it working. I thought it best to leave it in there in case someone else with more patience can figure it out, but I'll remove it if you don't want something fully functional in there.
> 
> Perhaps the right thing to do. You can leave a note if someone wants it.

I had it in the commit message, but in the new version, I added a comment in both the files to denote the issue

> > I'm not sure how to best handle the raspberrypi.conf being included in raspberrypi-cm.conf. I can move the common parts into the individual files and remove the include of raspberrypi.conf. Or, I can drop the qemu emulation of raspberrypi (raspi1ap) completely, and only have raspi0 emulated.
> 
> Like raspberrypi-cm3.conf, maybe make a copy.

I did this in the new version and rolled it into the commits where I added QB entries.

> > FYI, the Rasperry Pi machines that can be emulated are listed at https://www.qemu.org/docs/master/system/arm/raspi.html

jonmason:
Of note, with the last commit in the series, you should be able to run (in CI):
kas build kas/raspberrypi.yml:kas/testimage.yml
kas build kas/raspberrypi0.yml:kas/testimage.yml
kas build kas/raspberrypi2.yml:kas/testimage.yml

I was able to run all 3 successfully

PR #1595, opened 2026-04-29
linux-raspberrypi-6.12: Upgrade to 6.12.83

g-scott-murray:
This bumps to latest 6.12 on linux-rpi's rpi-6.12.y branch, matching linux-yocto in oe-core as of this posting.  The major motivator is to get the fix for building perf with glibc 2.43.


mast:
can we switch to 6.12.85 to address CVE-2026-31431 in this PR? it would be faster then creating new one.

g-scott-murray:
This has been superseded by @mast's 1597, dropping.

PR #1603, opened 2026-05-30
rpi-eeprom: devtool check-upgrade-status

vermaete:
before:
  rpi-eeprom                v2025.07.17-2712 UNKNOWN_BROKEN  None 05d94be4554ce44a057bfce8d0dd37d951703dab
now:
  rpi-eeprom                2025.07.17-2712 2026.05.11-2712 None e25fc5dcb8eb072eafb745cd546c3d9f73d102b5

### DIFF
--- a/recipes-bsp/rpi-eeprom/rpi-eeprom_git.bb
+++ b/recipes-bsp/rpi-eeprom/rpi-eeprom_git.bb
@@ -9,7 +9,7 @@ SRC_URI = " \
 "
 
 SRCREV = "2349daafacfb7a7abe2cfecf30a49ae837bdf2c6"
-PV = "v2025.07.17-2712"
+PV = "2025.07.17-2712"
 
 RDEPENDS:${PN} += " \
     coreutils \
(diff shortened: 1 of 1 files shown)


OldManYellsAtCloud:
Changing PV like this can cause "version goes backwards" problems for those who care about things like that (e.g. when updating package in place). Maybe it could use a `PE` also?

vermaete:
> `PE`

or is there a way to keep the 'v' as part of 'PV' with some regex magic?
Well, I could not find an example of it.

OldManYellsAtCloud:
Can you not just include the v in the capture also, like `^(?P<pver>v\d{4}\.\d{2}\.\d{2}-\d+)$`?

kraj:
> > `PE`
> 
> or is there a way to keep the 'v' as part of 'PV' with some regex magic? Well, I could not find an example of it.

I think PE bump is the familiar way to do it. Secondly, I wonder if its good idea to break the version bumps while not changing major version. Perhaps wait for next version bump

PR #1610, opened 2026-07-07
rpi-eeprom: update eeprom firmware to v2026.05.11-2712

embetrix:


OldManYellsAtCloud:
Since it is bumped, maybe https://github.com/agherzan/meta-raspberrypi/pull/1603 could be also incorporated? Or do you want to handle them as two PR after each other? @kraj @vermaete

kraj:
> Since it is bumped, maybe #1603 could be also incorporated? Or do you want to handle them as two PR after each other? @kraj @vermaete

I think that PR should be followed up separately. Since it still needs to have PE bumped.

PR #1611, opened 2026-07-17
rpi-config: enable RPi 3 x64 audio support only with 'ENABLE_AUDIO_BCM2835'

AdrianDC:
By enabling the audio module automatically only on RPi 3 x64 machines, the PWM* lines may get reused for audio purposes unwillingly, introducing a disparity with other RPi boards using the same sources

Create 'ENABLE_AUDIO' build constant to enable the audio support

### DIFF
--- a/docs/extra-build-config.md
+++ b/docs/extra-build-config.md
@@ -398,6 +398,13 @@ discovered via all 1-Wire busses check the interface with this command
 
 `ls /sys/bus/w1/devices/`
 
+## Enable audio BCM2835 sound support
+
+Audio driver for BCM2835 depends on the 'snd_bcm2835' module.
+To enable the audio support explicitly set it in `local.conf`
+
+    ENABLE_AUDIO_BCM2835 = "1"
+
 ## Manual additions to config.txt
 
 The `RPI_EXTRA_CONFIG` variable can be used to manually add additional lines to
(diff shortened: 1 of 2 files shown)


docs/extra-build-config.md
```diff
@@ -398,6 +398,13 @@ discovered via all 1-Wire busses check the interface with this command
 
 `ls /sys/bus/w1/devices/`
 
+## Enable audio sound support
+
+Audio driver for BCM2835 depends on the 'snd_bcm2835' module.
+To enable the audio support explicitly set it in `local.conf`
+
+    ENABLE_AUDIO = "1"
```
kraj: is this specific to rpi3? if so perhaps the option should be tailored to just rpi3

recipes-bsp/bootfiles/rpi-config_git.bb
```diff
@@ -200,7 +200,7 @@ do_deploy() {
         esac
     fi
 
-    if [ "${ENABLE_UART}"  = "1" ] && [ "${MACHINE}" = "raspberrypi-cm5-io-board" ]; then
+    if [ "${ENABLE_UART}" = "1" ] && [ "${MACHINE}" = "raspberrypi-cm5-io-board" ]; then
```
kraj, line 203: This should be a separate commit.

AdrianDC: Done :+1: !

docs/extra-build-config.md
AdrianDC: Hello Khem !

Actually you introduced these changes in 2017 for the RPi 3 x64 and it stayed limited to it (bb51049ebe761ac2806fa0fa492c586976632011),
I initially wanted to keep the new variable name generic like the others for future use on other audio parts.

Renamed it to `ENABLE_AUDIO_BCM2835`, let me know if you have other suggestions :+1:.

PR #1275, opened 2024-02-02
rpi-base: Add missing hifiberry overlay

pascalhuerst:
There is anoter hifiberry hat, for which there is currently no overlay added to the build, so let's add it:

hifiberry-dacplusadcpro.dtbo

**- What I did**

**- How I did it**

Issue #1168, opened 2023-04-18
Lacks support for Hifiberry Dac+ADC Pro


valentinlageard:
**Description**

Hi, I'm trying to build an image containing hifiberry-dacplusadcpro.dtbo.
I've seen that it is not included in RPI_KERNEL_DEVICE_TREE_OVERLAYS `conf/machine/include/rpi-base.inc`. 
I'm in the process of learning yocto and I'm not entirely sure how to append the overlay I need to the previously mentioned variable. I've tried to find information on how to change variables in .inc files to no avail.
I considered cloning and patching your layer but if I understand correctly, it is considered bad practice to clone and modify layers you depend on.
Could you point me to the right direction in order to have my custom layer adding support for the DAC + ADC Pro by appending to your layer ?
Or even, better, add support directly in your layer ?

Thanks and cheers,
Val.

valentinlageard:
Hi, I tried simply manually adding `overlays/hifiberry-dacplusadcpro.dtbo` to the `rpi-base.inc` file and the device tree overlay is added to the image and works fine.

Still, I don't understand how if I should do a pull request or find a way to override this in my own layer. If option 2 is better, could you point me in the right direction ?

Val.

JDavidsson:
Hi, 

It was added in https://github.com/agherzan/meta-raspberrypi/commit/fc34bc3b86f9eb42296618a3ceef82f276c472f3 (https://github.com/agherzan/meta-raspberrypi/pull/1275) and is present in kernels 6.1, 6.6, 6.12 and 6.18.

PR #1620, opened 2026-09-02
Hifiberry overlays update

JDavidsson:
- What I did

Added five HiFiBerry device tree overlays to RPI_KERNEL_DEVICETREE_OVERLAYS in conf/machine/include/rpi-base.inc, as two commits.

hifiberry-dacplus-std and hifiberry-dacplus-pro:

HiFiBerry split the DAC+ overlay in two, because the Pi 5's driver infrastructure no longer lets a card reconfigure its clocking at runtime — the interface has to be declared master or slave in the overlay itself. A DAC+ Standard, Amp2 or Amp4 now needs -std, a DAC+ Pro or DAC2 Pro needs -pro. Neither is deployed by this layer, while the superseded hifiberry-dacplus still is — so the overlay a current DAC+ owner needs is the one that's missing, and every older guide points at the one that still loads.

- kernel commit: https://github.com/raspberrypi/linux/commit/9e31e8ce44ef
- HiFiBerry's explanation: https://www.hifiberry.com/blog/changes-in-hifiberry-drivers/

The other three — hifiberry-adc, hifiberry-adc8x and hifiberry-dac8x — are built by the kernel but not deployed by this layer.

In both cases the failure is silent: a dtoverlay= line naming a .dtbo that isn't on the boot partition doesn't stop the boot and prints nothing an ordinary user will look for. The card simply never appears.

- How I did it

Five lines added to the existing list.

hifiberry-studio-dac8x and -pro are deliberately excluded — added to the kernel in ba7ae2c9, absent from the kernel the 6.6 recipe pins. Verified on a build: with PREFERRED_VERSION_linux-raspberrypi = "6.6.%" the five overlays added here compile, while adding hifiberry-studio-dac8x on top fails do_compile with "No rule to make target 'arch/arm/boot/dts/overlays/hifiberry-studio-dac8x.dtbo'".

After reading #1168 I checked the same thing for 6.1: hifiberry-adc and hifiberry-adc8x are not in the kernel linux-raspberrypi_6.1.bb pins (fbd8b3fa), so they would break a 6.1 build the way imx708 broke 5.10 in #1158. Happy to drop those two if you'd prefer — the other three are present in all four kernels.

Btw, #1168 asked for hifiberry-dacplusadcpro. It was added in fc34bc3 (#1275) and is present in 6.1, 6.6, 6.12 and 6.18.

### DIFF
--- a/conf/machine/include/rpi-base.inc
+++ b/conf/machine/include/rpi-base.inc
@@ -28,11 +28,14 @@ RPI_KERNEL_DEVICETREE_OVERLAYS ?= " \
     overlays/gpio-key.dtbo \
     overlays/gpio-poweroff.dtbo \
     overlays/gpio-shutdown.dtbo \
+    overlays/hifiberry-adc.dtbo \
+    overlays/hifiberry-adc8x.dtbo \
     overlays/hifiberry-amp.dtbo \
     overlays/hifiberry-amp100.dtbo \
     overlays/hifiberry-amp3.dtbo \
     overlays/hifiberry-amp4pro.dtbo \
     overlays/hifiberry-dac.dtbo \
+    overlays/hifiberry-dac8x.dtbo \
     overlays/hifiberry-dacplus-pro.dtbo \
     overlays/hifiberry-dacplus-std.dtbo \
     overlays/hifiberry-dacplus.dtbo \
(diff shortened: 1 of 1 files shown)


kraj (approved):
LGTM

JDavidsson:
Seems to be a timeout on one of the test jobs?

